PR #7, opened 2014-09-04
Fixes two bugs in the SeLion-Code-Generator

renelux:
Code generator went through all .yaml in scr/main/resources which will cause errors if user has .yaml files that are not compatible with the SeLion-Code-Generator. SeLion-Code-Generator will now stay in the GUIData folder as described by the documentation.

fixed: .yaml files in the GUIData root folder can now be parsed correctly

### DIFF
--- a/codegen/src/main/java/com/paypal/selion/plugins/CodeGeneratorMojo.java
+++ b/codegen/src/main/java/com/paypal/selion/plugins/CodeGeneratorMojo.java
@@ -50,6 +50,13 @@ public class CodeGeneratorMojo extends AbstractMojo {
      * @parameter expression="${selion-code-generator.basePackage}" default-value="com.paypal.selion.testcomponents"
      */
     private String basePackage;
+    
+    /**
+     * Represents the base folder used for reading yaml files.
+     * 
+     * @parameter expression="${selion-code-generator.baseFolder}" default-value="GUIData"
+     */
+    private String baseFolder;
 
     /**
      * List of "domains" to exclude during code generation.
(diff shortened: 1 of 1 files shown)


codegen/src/main/java/com/paypal/selion/plugins/CodeGeneratorMojo.java
```diff
@@ -110,8 +110,13 @@ private String resourcesDir() {
 
     private String pathToFolder(File file) {
         String path = file.getAbsolutePath();
-        return path.substring(path.indexOf("GUIData") + "GUIData".length() + 1, path.lastIndexOf(File.separator))
-                .trim();
+        try {
+        	return path.substring(path.indexOf("GUIData") + "GUIData".length() + 1, path.lastIndexOf(File.separator))
```
mach6: do I detect a TAB character here?

codegen/src/main/java/com/paypal/selion/plugins/CodeGeneratorMojo.java
```diff
@@ -244,6 +249,7 @@ private boolean generateJavaCode(File baseFile, File dataFile, File extendedFile
      * @return List<File>
      */
     private List<File> loadFiles(File workingDir) {
+    	System.out.println(workingDir.getAbsolutePath());
```
mach6: Is this System.out required?

renelux:
I remember that this was configurable before but I think it was removed some time back? What would be the best way to provide this setting to the plugin?

sebady:
I would consider making the yaml base path configurable in the plugin's configuration and not hard code it in the implementation.  See File/Directory handling here
http://maven.apache.org/guides/plugin/guide-java-plugin-development.html

mach6:
@renelux I went through the history. It was never configurable that I see.

renelux:
ok, I will update the pull-request tomorrow to make it configurable.

renelux:
I noticed that I forgot to added comment, I have pushed the changes making the Yaml folder configurable :)

mach6:
@renelux  I made minor tweaks to your merge commit and a javadoc comment then merged this PR

PR #23, opened 2014-09-16
Added support for event listeners for actions done on elements provided by SeLion.

renelux:
Adds support to SeLion to trigger events before and after we perform actions on SeLion elements. This will help us to make SeLion more extendable.

The following events are supported:

Click
Type
Check
Uncheck
Submit
Select
Deselect


mach6:
@fakrudeen78 @krmahadevan @sebady I'm inclined to merge this. Do any of you have feedback on this PR? I know there was talk about accomplishing this a different way. Do we still want to pursue that?

sebady:
This approach for supporting event listeners looks reasonable to me.  I'm not aware of an alternative under consideration/discussion. I would merge it.

krmahadevan:
I haven't yet had the chance to revisit this yet. Been swamped with stuff. I would want to look at this again since I was involved in the original discussions and first set of review on this.
- Cheers
  Krishnan.

Sent from a mobile.
So if you noticed any typos you know whom to blame 😊

-------- Original message --------

krmahadevan:
@mach6 
@fakrudeen78

Guys I am fine with this implementation being merged.

But we need to take care of the below mentioned trivial things before the code hits `develop`
- [x] In `com.paypal.selion.platform.grid.AbstractTestSession.getListeners()` We can consider renaming this to `getSeLionEventListeners()` to make sure the intent of this method is clear.
- [x] In `com.paypal.selion.platform.grid.ScreenShotRemoteWebDriver.register(SeLionElementEventListener)` - 
  Please add a Null Check here, because this is attempting to add an element to the arraylist and I dont think array list does a null check (as the List contract states).
- [x] In `com.paypal.selion.platform.html.support.events.AbstractSeLionElementEventListener` This class should be an abstract class rather than a concrete class so that it is in sync with what `org.openqa.selenium.support.events.AbstractWebDriverEventListener` works as in principle.
- [x] In the package `com.paypal.selion.platform.html.support.events` all the interfaces need some basic javadocs which convey the intent of these marker interfaces.
- [x] We need to clean up the un-used imports in `com.paypal.selion.testcomponents.BasicPageImplTest`.

mach6:
@krmahadevan @fakrudeen78 @renelux  I've pulled this contribution into the branch https://github.com/paypal/SeLion/tree/renelux-patch-add-events so that we can address the review comments. Furthermore, I have rebased it off of the latest https://github.com/paypal/SeLion/tree/develop  branch

mach6:
I also changed all references of `SeLionElementEventListener` to `ElementEventListener`. I don't see the need for the use of the word SeLion since 1. There is no ElementEventListener's in Selenium code (that I see) and 2. the word selion is already in the package.

mach6:
This is now merged to develop. Issue #38 opened to track the documentation needs for this PR.

Issue #15, opened 2014-09-09
Documentation - footer does not always stay on the bottom


mach6:
http://paypal.github.io/SeLion/html/documentation.html  

The footer on this section of the site does not always stay on the bottom as it should.

sundaramrajendran:
@mach6 

Is this specific to any browser or any page for this issue. When I try to load the page footer is getting displayed and once content is loaded footer is moved to bottom of the page. Is this issue?

mach6:
@sundaramrajendran the issue is on any browser, from what I see. The problem is the footer does not always align to the absolute bottom of the viewport/browser window -- like it does on the front page -- http://selion.io  ... to see the issue you need to have a viewport/browser with a large vertical area or load a page with a small amount of text.   The point you mention of the footer displaying and then moving directionally towards the bottom after the content is loaded is a secondary issue which can easily be fixed with some javascript which changes visibility from hidden to visible on document load.

sundaramrajendran:
@mach6 I am working on this issue

mach6:
PR merged to `gh-pages`

PR #29, opened 2014-09-25
Fix (issue #15) to remove the unwanted whitespace below the footer

sundaramrajendran:
This fix has been test with FireFox, Chrome, IE and Safari browsers

### DIFF
--- a/css/selion.css
+++ b/css/selion.css
@@ -3,11 +3,14 @@
 
 html {
   position: relative;
-  min-height: 100%;
+  height: 100%;
 }
 
 body {
+  position: absolute;
   color: #5a5a5a;
+  min-height: 100%;
+  min-width: 100%;
 }
 
 h1 {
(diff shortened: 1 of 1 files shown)


sundaramrajendran:
@mach6 Addressed all the review comments.

mach6:
@sundaramrajendran  The footer behavior in the PR for the documentation page is not what I had in mind. 

What I'm looking for is a footer that is always at the absolute bottom of the page -- like the home page -- not a "sticky" footer which is at the bottom of the view port.. In the later, the content above the footer is scrolled in the container div, and the footer remains.. In the former, the content above the footer is scrolled with the entire page. The footer will only be visible when the user scrolls to the bottom of the page.

sundaramrajendran:
@mach6 I hope latest commit meets your requirement

mach6:
@sundaramrajendran looks really good and such a simple fix too! very nice.

PR #40, opened 2014-10-20
Added the code to specify the parameter to forceReinstall the Selendroid Server and AUT

rkpaypal:
Usage: 

User can specify the forceReinstall parameter in the TestNG suite file as true to forceReinstall the Selendroid Server and AUT. This code helps for fresh reinstall of Selendroid Server ans AUT as sometimes there might resign issues with Selendroid Server and AUT not being resigned with same debug key and the test the might to start up.

Regards
Rajani


mach6:
@rkpaypal You still have a problem with your commit messages and furthermore now the commits need squashing.

Issue #45, opened 2014-10-21
waitUntilElementIsXXX functions do not handle piped locators properly


azaytsev:
waitUntilElementIsXXX methods timeout if a piped element locator is passed

mach6:
closing. this is merged to develop

PR #46, opened 2014-10-21
Fixed issues #45 and added a supporting unit test

azaytsev:
Fixed issue #45 and added a supporting unit test

### DIFF
--- a/client/src/test/java/com/paypal/selion/platform/grid/SeLionSeleniumTest.java
+++ b/client/src/test/java/com/paypal/selion/platform/grid/SeLionSeleniumTest.java
@@ -15,109 +15,97 @@
 
 package com.paypal.selion.platform.grid;
 
-import static org.testng.Assert.assertTrue;
-import static org.testng.Assert.fail;
-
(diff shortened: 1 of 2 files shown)


client/src/test/java/com/paypal/selion/platform/grid/SeLionSeleniumTest.java
```diff
@@ -101,6 +102,18 @@ public void testWaitUntilElementPresentPos() {
...
     }
 
+    @Test(groups = { "browser-tests" })
+    @WebTest
+    public void testWasitUntilElementPipedLocator() {
+        Grid.driver().get(url);
+        try {
+            WebDriverWaitUtils.waitUntilElementIsPresent(pipedLocator);
+            assertTrue(true);
+        } catch (WaitTimedOutException e) {
```
krmahadevan: @azaytsev 
I believe that TestNG by default fails a test method if the Test method generates an exception. So do we really need to wrap this within a try..catch block ?

mach6:
this is merged to develop.

Issue #30, opened 2014-10-01
SeLion logged content is written to ios-driver.log when running tests against ios-driver


mach6:
ios-logging.properties installs a FileHandler on the root logger which outputs ALL events to ios-driver.log
This is a pretty invasive assumption for downstream applications such as SeLion.

mach6:
merged to `develop`.

PR #55, opened 2014-11-20
Passing a config skipLoggingConfiguration flag to ios-driver to skip ios...

azaytsev:
...-driver logger configurations. See issue #30.
- This change should not be merged until https://github.com/ios-driver/ios-driver/issues/322 has been merged


mach6:
merged to `develop`

PR #59, opened 2014-12-07
Split app name and version in Capabilities

binhvu:
To enhance capability matching and select the latest version of the app to test if the
app version is not provided by the users.

Signed-off-by: Binh Vu bvu@paypal.com

### DIFF
--- a/client/src/main/java/com/paypal/selion/configuration/Config.java
+++ b/client/src/main/java/com/paypal/selion/configuration/Config.java
@@ -475,6 +475,19 @@ public final class Config {
         SELENIUM_NATIVE_APP_NAME("appName", "", false),
 
         /**
+         * This parameter represents the version of the app that is to be spawned. There are two way to specified the
+         * version of the native app.<br>
+         * Specify app version as part of the appName as:
+         * <pre>
+         * appName = &quot;Safari:7.0&quot;
+         * </pre>
+         * or by itself as:
+         * <pre>
+         * appVersion = &quot;7.0&quot;
+         */
+        SELENIUM_NATIVE_APP_VERSION("appVersion", "", false),
+
+        /**
          * This parameter represents the language to be used. By default it is always <code>English</code> represented
          * as <code>en</code>
          */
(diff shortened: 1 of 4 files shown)


binhvu:
This build will fail until my selendroid PR is merge.

client/src/main/java/com/paypal/selion/configuration/Config.java
```diff
@@ -475,6 +475,19 @@ public static synchronized void setConfigProperty(ConfigProperty configProperty,
...
+         * version of the native app.<br>
+         * Specify app version as part of the appName as:
+         * <pre>
+         * appName = &quot;Safari:7.0&quot;
+         * </pre>
+         * or by itself as:
+         * <pre>
+         * appVersion = &quot;7.0&quot;
+         */
+        SELENIUM_NATIVE_APP_VERSION("appVersion", "", false),
```
mach6, line 488: I don't think calling this `SELENIUM_NATIVE_APP_VERSION` make sense. As coded, only the selendroid integration will make use of it. 

@azaytsev Could our ios-driver integration make use of this config param?

Also FYI, I realize you followed the approach used by `SELENIUM_NATIVE_APP_NAME` here but it's case I'm not sure it makes sense to have as a Config variable. My reason for saying this -- I just don't realistically see a world where a single Native app Test will work across multiple apps without significant `if` .. `else` constructs in the Test. Certainly, the PayPal native app Test won't work for the eBay native app and vice-versa.. Thoughts on this?

mach6, line 488: @azaytsev I answered my own question. 

see `IOSCapabilities.BUNDLE_VERSION` and `IOSCapabilities.BUNDLE_SHORT_VERSION`

Can you help @binhvu test with these `IOSCapabilities` set with the value of `appVersion` per this PR?

azaytsev, line 488: @mach6 I'll sync up with @binhvu tomorrow and help him figure this out.

PR #60, opened 2014-12-11
Add functionality for a user to hover an element.

renelux:
Adds functionality for a user to hover an element. The user can also provide expected objects that should be visible after hovering the element.

An screenshot gets automatically take once we hovered to the element. We also trigger events just like other actions we perform on elements.

The hover method should be useful for handling websites with dropdown menus.

### DIFF
--- a/client/src/main/java/com/paypal/selion/platform/html/AbstractElement.java
+++ b/client/src/main/java/com/paypal/selion/platform/html/AbstractElement.java
@@ -31,6 +31,7 @@ import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
+import org.openqa.selenium.interactions.Actions;
 import org.openqa.selenium.remote.RemoteWebElement;
 import org.openqa.selenium.support.ui.ExpectedCondition;
 import org.openqa.selenium.support.ui.WebDriverWait;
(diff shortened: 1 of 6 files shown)


krmahadevan:
@renelux 

This implementation would only be useful in cases wherein a hover would trigger something (for e.g., a tool tip). I am not quite sure how this implementation is going to help people who need to deal with drop down menus that get shown when you hover over something and a user would need to interact with one of the menu items that get shown after the hover action ?

renelux:
@krmahadevan you are correct it only works if the hover triggers something. So if you dropdown menu gets triggered by a click then it won't help. 

If you have a look at the following page http://cssmenumaker.com/menu/flat-jquery-responsive-menu
then need to hover products first before you can click Product 1. So with the new hover method you can accomplish that with the following code.

``` java
Grid.driver().get("http://cssmenumaker.com/menu/flat-jquery-responsive-menu");

Link link = new Link("css=#cssmenu .active.has-sub");
Link subLink = new Link("css=li.has-sub a");

link.hover(subLink);
subLink.click();
```

And yes it is also very helpful for things like tooltips :)

krmahadevan:
@renelux 
Hmm.. I was of the impression that in order to work with drop down menus, the action of moving to the main menu item (which would cause the drop down menu to appear), the action of finding the item in the menu of the drop down menu and the action of performing a click on the menu item should be done in one shot as part of the Same `Actions` class.
The reason being, only when the mouse is moved to the main menu, would the menu items be even visible on the DOM (Selenium WebDriver is strict on these things and as you know it simulates the actual user behavior). 
But this implementation breaks that chain by just causing the mouse to move the menu. 

Have you tried this with multiple places wherein these kind of menu items are present ?

renelux:
I did not know that it is suggested by Selenium to chain actions like this. If that is the case then yes we should not be using it for such actions (although it does work).

Do you think it still adds value for things like tooltip? We have a few cases where we need to trigger the tooltips in our TC and create a screenshot for LQA verifications.

krmahadevan:
@renelux 

I think I will play around more on this and update this pull request with my findings perhaps tomorrow or over the weekend.

mach6:
assigned to @krmahadevan for follow-up

krmahadevan:
@renelux 
Thanks for providing this Pull Request. I have merged this. Looks like WebDriver leaves the mouse in the last web element and as such actions such as navigating a multi level drop down menu can be done using the `hover` action. So this would be definitely a value add to SeLion.

PR #61, opened 2014-12-11
Fixes issue where if we call isElementPresent and there are no page titles defined it will throw an NullPointException

renelux:
This PR address an issue that if we call isElementPresent on a element of which the page yaml has no pageTitle defined it will throw an Exception.

Issue comes from the following line of code the method add extra info when a element can't be located. But on the following line [AbstractElement.java#L148](https://github.com/paypal/SeLion/blob/develop/client/src/main/java/com/paypal/selion/platform/html/AbstractElement.java#L148). we verify if the elements parentPage is loaded based on pageTitle. Is that something that we should be doing? Isn't it up to the user to ensure that they are on the correct page?


krmahadevan:
@renelux 
I will take a look at this tomorrow.

renelux:
@mach6 I updated the code according to your suggestions.

mach6:
:+1: Looks good to me. I'll merge soon.

krmahadevan:
@mach6 
I went ahead and merged this, since I am working on merging other changes as well.Hope it is fine with you :)

Issue #53, opened 2014-11-14
@MobileTest iOS not honoring locale and language settings


renelux:
Please see the example below if I run this test case the simulator is still being launched with language=en and locale=en_US.

``` java
package com.symbio.test;

import java.util.List;

import org.openqa.selenium.By;
import org.openqa.selenium.WebElement;
import org.testng.Assert;
import org.testng.annotations.Test;

import com.paypal.selion.annotations.MobileTest;
import com.paypal.selion.platform.grid.Grid;
import com.paypal.selion.reports.runtime.MobileReporter;

public class IOS_Multi_Locale {

    @MobileTest(appName = "InternationalMountains", language="fr", locale="fr_FR")
    @Test
    public void testMethod() throws InterruptedException {
        MobileReporter.log("My Screenshot 1", true);
        List cells = Grid.iOSDriver().findElements(By.className("UIATableCell"));
        Assert.assertEquals(9, cells.size());

        // get the 1st mountain
        WebElement first = (WebElement) cells.get(0);
        first.click();
        Thread.sleep(10 * 1000);

        // take a screenshot using the normal selenium api.
        MobileReporter.log("My Screenshot 2", true);

        // access the content
        By selector = By.xpath("//UIAStaticText[contains(@name,'climbed')]");
        WebElement text = Grid.iOSDriver().findElement(selector);
        System.out.println(text.getAttribute("name"));
    }

}
```

azaytsev:
@renelux - I just ran your test on my machine and it's definitely setting the locale and language to French. In fact, it fails because of the following line in your test, since the text is now in French. 
        `By selector = By.xpath("//UIAStaticText[contains(@name,'climbed')]");`

![internationmountains_scr1](https://cloud.githubusercontent.com/assets/4985640/5076677/bd920cee-6e51-11e4-9ccf-4e3147a06203.png)
![internationmountains_scr2 png](https://cloud.githubusercontent.com/assets/4985640/5076683/c4e18e52-6e51-11e4-858c-df47d1cd71a6.png)

renelux:
@azaytsev I am still running in to the same issue. Can you please tell me what Xcode you are using? And what simulator. I am on the latest version for both and I am not sure if it is supported correctly.

Is there a way to set the simulator that I want to use?

azaytsev:
@renelux, I was able to test with both XCode 6.0.1 / IOS 8.0 simulator and XCode 5.1.1 / iOS version 7.1

[...]

PR #62, opened 2014-12-15
Adding capability to specify sdk version / device variation via Webtest ...

azaytsev:
Addressing ticket #53. Adding capabilities to allow a use to specify sdk / device variation

### DIFF
--- a/client/src/main/java/com/paypal/selion/platform/grid/MobileTestSession.java
+++ b/client/src/main/java/com/paypal/selion/platform/grid/MobileTestSession.java
@@ -39,6 +39,8 @@ public class MobileTestSession extends AbstractTestSession {
     private String appLanguage;
     private String appLocale;
     private String deviceSerial;
+    private String deviceType;
+    private String platformVersion;
 
     MobileTestSession() {
         super();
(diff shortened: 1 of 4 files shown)


mach6:
Does this PR include the changes @binhvu needs?

binhvu:
@mach6
Yes. It has all the common parameters that selendroid needed. I will need to work with selendroid first then come back to work on SeLion.

client/src/main/java/com/paypal/selion/platform/grid/MobileTestSession.java
```diff
@@ -80,15 +82,27 @@ public WebDriverPlatform getPlatform() {
...
     }
 
+    public String getPlatformVersion() {
+        return platformVersion;
+    }
+
     public String getDevice() {
-        if (device.contains("android") || device.equalsIgnoreCase("iphone") || device.equalsIgnoreCase("ipad")) {
+        if (device.contains("android") || device.contains("iphone") || device.contains("ipad")) {
+            if (StringUtils.contains(device, ":")) {
```
krmahadevan, line 91: Do we really need to use `StringUtils.contains` and `StringUtils.split` ? Can we not just use `device.contains` and `device.split` here ? I believe `StringUtils` basically shields you for null values, but if that were the case then the parent if condition itself would be throwing a `NullPointerException` because we have not shielded the code there.

client/src/main/java/com/paypal/selion/platform/grid/browsercapabilities/GenericCapabilitiesBuilder.java
```diff
@@ -37,17 +41,33 @@ public DesiredCapabilities getCapabilities(DesiredCapabilities capabilities) {
...
         DesiredCapabilities caps = null;
-        if (this.deviceType.equalsIgnoreCase("ipad") || this.deviceType.equalsIgnoreCase("iphone")) {
+
+        if (this.device.contains("ipad") || this.device.contains("iphone")) {
             caps = new IOSCapabilities();
-            caps.setCapability(IOSCapabilities.DEVICE, this.deviceType);
+
+            String appVersion = null;
+            if (StringUtils.isNotBlank(appName) && StringUtils.contains(this.appName, ":")) {
+                appVersion = StringUtils.split(this.appName, ":")[1];
```
krmahadevan, line 50: I think we should add an edit check, which basically checks if the split resulted in an array of size 2 or more else this can throw `ArrayIndexOutOfBoundsException` for an app name which resembles like `foo:` no ?

azaytsev, line 50: That's a good idea. I'll add one in

client/src/main/java/com/paypal/selion/platform/grid/MobileTestSession.java
azaytsev, line 91: @krmahadevan good point. I'll add an NPE check and switch it to using device.contains/device.split

Issue #57, opened 2014-12-01
Add an ability to specify SDK version and/or device variation when launching ios-driver


azaytsev:
Add an ability to specify sdk version and/or device variation via Webtest annotation when launching and ios-driver test

PR #67, opened 2015-01-12
Adding capability to specify sdk version / device variation via Webtest ...

binhvu:
...annotations. #57

Providing a way to pick a specific device to test when there are
more than one devices with the same SDK version.

For IOS this can be iPhone 5/5S, iPhone 6/Plus...
For android this can be Nexus 4, Galaxy 3S...

Signed-off-by: Binh Vu bvu@paypal.com


mach6:
this is merged to the feature branch

PR #78, opened 2015-03-06
Clean up references to ebay and ebay servers in reporter files.

ILikeToNguyen:
Renamed euqe.css and euqeb.js to htmlreporter.css and htmlreporter.js to be more clear what these files are used for.
Removed duplicate css entries in euqe.css


client/src/main/resources/templates/bluefin-js-css/euqe.css
```diff
@@ -1,1249 +0,0 @@
-.yuk_table td.clickedOnce {
```
ILikeToNguyen, line 1: FYI i tried to do the proper git commands for a file rename and edit the file but I think because my change ripped out half the files since there was a huge duplicate section, github thinks its a different file.  

I can do 2 commits where i change the file name, do a commit, then edit the file and it will keep the tracking correct.  Should I do that?  Same applies to euqe.js

mach6:
@ILikeToNguyen made some tweaks and merged this.

PR #79, opened 2015-03-11
Add support for specifying additional DesiredCapabilities for Mobile Tests

ILikeToNguyen:
- Add support for 'additionalCapabilities' annotation parameter to @MobileTest so that users can set additional DesiredCapabilities.
- Add support for boolean DesiredCapabilities.
- Allow colons to be in the value part of name:value

NOTE: I changed the order of priority.  DesiredCapbilities from annotation param now takes priority over DesiredCapabilites from ExtendedConfig.  Makes mores sense for the lowest level to take precedence in value over higher level.  This may break existing user's if they specified the same desiredcapability in ExtendedConfig and in annotation parameter.

### DIFF
--- a/client/src/main/java/com/paypal/selion/annotations/MobileTest.java
+++ b/client/src/main/java/com/paypal/selion/annotations/MobileTest.java
@@ -79,4 +79,17 @@ public @interface MobileTest {
      */
     String deviceType() default "";
 
+    /**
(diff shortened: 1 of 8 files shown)


mach6:
@ILikeToNguyen minor tweaks made and merged this

PR #80, opened 2015-03-16
RuntimeReporter has skipped tests also marked as running when test run is complete

ILikeToNguyen:
Runtime Reporter: skipped tests need to be removed from running tests list so that the report doesn't leave skipped tests marked as running.


mach6:
@sebady verify and merge please. should be quick

sebady:
Merged

PR #82, opened 2015-03-20
Set BrowserName in DesiredCapabilities to android for Webview and set to selendroid for native app

renniechan:
BLUEFIN-2023 set BrowserName in DesiredCapabilities to android for Webview and set to selendroid for native app.


mach6:
@renniechan this is merged. -- I did change your commit message a bit

Issue #84, opened 2015-03-24
Code-Generator produces code which will not compile when PageYaml document resides in src/main/resources/GUIData


mach6:
- [x] If you put a PageYaml document in the top-level folder under src/main/resources/GUIData, the code-generator will produce a corresponding class with and invalid `package` declaration -- it will end with a `.`
- [x] Also, the site documentation http://paypal.github.io/SeLion/html/documentation.html#what-is-code-generator, bullet 2 under "Put the SeLion Code Generator in Action" is mis-leading. The file, once created, should reside in ANY directory under src/main/resources/GUIData ... It does not have to be placed in the top-level of this location.

ysecaran:
@mach6 I am looking into this

mach6:
fix is merged

PR #86, opened 2015-04-01
Fix for #84 and misc clean-up

mach6:
1. Allow for PageYaml to reside in top-level GUIData directory.
2. Remove PageYamlException.java which was not needed given it's current
   use.


mach6:
@krmahadevan assigned this to you, if you have time. Thanks!

mach6:
@fakrudeen78 this PR has been updated to address your comment

codegen/src/main/java/com/paypal/selion/plugins/CodeGeneratorMojo.java
```diff
@@ -124,6 +125,11 @@ private String pathToFolder(File file) {
      */
     public void execute() throws MojoExecutionException {
 
+        // iff basePackage is null, empty, or just whitespace reset it to the default value
+        basePackage = StringUtils.defaultString(basePackage, "com.paypal.selion.testcomponents");
```
fakrudeen78, line 129: It is better to raise exception when both basePackage and folder is empty. 
he need to provide at least one [basePackage or folder ].

with the current change if the user purposefully made basePackage empty and given detailed folder structure, here we are appending "com.paypal.selion.testcomponents" even though he provided folder structure.

mach6, line 129: @fakrudeen78 I see your point but I don't agree with the exception approach. I'd rather resort to convention. So, at a minimum a user will need to define a `basePackage` at least one level deep (such as `com`) or accept the default.

mach6, line 129: @fakrudeen78 How about 

``` java
if (basePackage.isEmpty() && folder.isEmpty()) {
  basePackage = "default";
}
```

I believe this approach will meet my ask w.r.t convention, allows for the user to define a detailed folder structure under `baseFolder` and leave `basePackage` empty, accommodates any files directly in the `baseFolder`, and avoids any exception.

krmahadevan, line 129: @mach6 

If the user purposefully or by mistake ended up overriding the value of `basePackage` with an empty value [ Can happen in scenarios wherein the user decides to use a Maven property to define this value, but forgets to provide a value for the Maven property via a JVM argument ] and if the user intentionally had all of his/her Yaml files under `GUIData` then shouldn't we be letting the user that some did go wrong rather than us trying to fix the problem on our own with a `default` value and thus going against what the user might have been expecting ? 
In the scenario that I explained, the application behaved in a way which the user didn't expect i.e., he thought he had provided a value for the `basePackage` via a Maven property and yet he/she sees the package name beginning with `default`, they would be misguided thinking that this is a bug in SeLion whereas the actual problem resided in their own test project ? Wouldn't that be an anomaly ?

mach6, line 129: @krmahadevan put a java file in `src/main/java` in a maven project. don't give it a `package` declaration. See what "package" the class ends up in -- `(default package)`.  Certainly `(default package)` != `package default;`  However, I still feel some sort of convention/solution is better than an exception here. Maven users are used to it applying conventions. 

We could also just change the velocity template to omit the `package` line when `basePackage` and `folder` are empty.

krmahadevan, line 129: @mach6 
The behavior kind of seems to be different between IDEs. eclipse shows it under a package named `default package` and Intellij doesn't do any of that. In either cases, the class definitely doesn't include a `package` declaration.

This makes sense to me here because I didn't provide a package name and the IDE defaulted to whatever was supposed to be its known default. But in our case, we are explicitly overriding the values with a value which is null or empty. So shouldn't the system just default to the edit checks in place to detect this rather than perform a recovery for the user ?
We do this in all other places such as our `Config` class wherein if the user explicitly overrides the browser flavor to an empty value, we try to use to and then throw an error no ? Why this change of behavior here ?

Another example of wherein we are going against the Maven conventions.
The maven surefire plugin lets us define the suite xml location via a properties. If I explicitly default it to an empty value, I guess I see a `NullPointerException` or something on those lines. 

So IMO we should use defaults if no values were provided. If values were provided we attempt to validate the values. If the validation fails, throw an error and bail out. If the values were valid, we just use them and move ahead.

mach6, line 129: > We do this in all other places such as our Config class wherein if the user explicitly overrides the browser flavor to an empty value, we try to use to and then throw an error no ? Why this change of behavior here ?

`Config` is not presented as or part of a maven plugin

> Another example of wherein we are going against the Maven conventions.
> The maven surefire plugin lets us define the suite xml location via a properties. If I explicitly default it to an empty value, I guess I see a NullPointerException or something on those lines.

I don't buy this argument. The `NullPointerException` in this case is because there is no acceptable default for surefire to figure out or convention that they could have coded (other than search the source tree for a testng suite file -- which seems overzealous) for it to fallback on. However, if the user defined a default `<property>` `<suiteXmlFile>` in the POM, that condition goes away too. 

> So IMO we should use defaults if no values were provided. If values were provided we attempt to validate the values.

agree, agree,

> If the validation fails, throw an error and bail out.

 disagree for this scenario.. Especially since omitting a `package` declaration in this situation solves the problem AND is more in line with what you would expect -- compared to the eclipse and intelij scenarios. So, I'm leaning in the direction that the right place to solve this is in the velocity template .. if `$package` is empty then do not write the  `package $package;` line

>  If the values were valid, we just use them and move ahead.

agree

mach6:
@fakrudeen78 @krmahadevan Please review and merge #88 which addresses this last piece of feedback regarding allowing for an empty `basePackage`

PR #88, opened 2015-04-07
Address possible issue in codegen if basePackage is empty

mach6:
With this change a user can specify `-Dselion-code-generator.basePackage=`
in the `mvn generate-sources` step. The resulting java files will have a
`package` line created entirely from the folder structure under the `baseFolder`
Furthermore, if there are files in `baseFolder`, they will end up in the
default package.

### DIFF
--- a/codegen/src/main/java/com/paypal/selion/plugins/CodeGeneratorMojo.java
+++ b/codegen/src/main/java/com/paypal/selion/plugins/CodeGeneratorMojo.java
@@ -79,7 +79,7 @@ public class CodeGeneratorMojo extends AbstractMojo {
     }
 
     public void setDetailedTextOutputLocation(File location) {
-        this.detailedTextOutputLocation = location;
+        detailedTextOutputLocation = location;
     }
 
     public void displayProjectInformation() {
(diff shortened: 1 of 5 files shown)


mach6:
@sebady Will you take a look at this please? @fakrudeen78 did not get a chance.

codegen/src/main/java/com/paypal/selion/plugins/CodeGeneratorMojo.java
```diff
@@ -165,9 +162,14 @@ public void execute() throws MojoExecutionException {
                     logger.info("Generating java file for YAML file [" + dataFile.getName() + "] in domain ["
                             + domain + "]");
                     
-                    String tempPackage = basePackage;
+                    String tempPackage = "";
+                    // add the basePackage, iff it contains a value
+                    if (!basePackage.isEmpty()) {
+                        tempPackage += basePackage;
```
sebady: The if test is not strictly necessary now.  The direct "tempPackage += basePackage;" now handles both cases.

mach6: Good point. Will fix this and merge.

mach6:
updated and merged.

codegen/src/main/java/com/paypal/selion/plugins/CodeGeneratorMojo.java
```diff
@@ -165,9 +162,14 @@ public void execute() throws MojoExecutionException {
...
+                    }
+                    // add the folder, iff it contains a value
                     if (!folder.isEmpty()) {
-                        tempPackage += "." + folder;
+                        tempPackage += tempPackage.isEmpty() ?  folder :  "." + folder;
                     }
                     helper.generateNewCode(dataFile, relativePath, tempPackage, domain);
```
sebady, line 170: so how are we handling an empty base package now?  Does a user get a valid error on empty package?

mach6:
no. there is no error message (and there should not be) on empty package.  we simply do not add a `package` line in the created java when the yaml file is in the `baseFolder` and `basePackage` is empty.. this results in a java class that is in the default package

When the yaml file is in a subdirectory under `baseFolder` and `basePackage` is empty, a package line is added to the created java class. the `package` line is the folder structure under `baseFolder`

PR #89, opened 2015-04-07
Fix getting ClassCastException on DesiredCapabilities to IOSCapabilities

ILikeToNguyen:
IOSDriver is the only driver that takes a non DesiredCapabilities
object.  Therefore we need to create the capabilities as an
IOSCapabilities instance, not a DesiredCapabilities instance.


mach6:
merged.

Issue #81, opened 2015-03-18
Code-Generator: Add support for custom elements


renelux:
We would like to add support for custom elements in the CodeGenerator. We would like to handle webpages that are build with ExtJS and if you look at for [example](http://dev.sencha.com/deploy/ext-4.0.0/examples/form/fieldcontainer.html) their drop down list. You will see that the default implementation is not going to work.

For us to be able to use our best practices for web automation we would like to have custom element in the CodeGenerator that can handle this kind of non default behavior.
# Proposed implementation:

Add a property to the Code-Generator which accepts a comma separated list of java classes. Based on the parameters we can add the imports and create getters and setters in the generated Pages. Example "com.rlux.test.elements.ExtJSSelectList,com.rlux.test.elements.ExtJSTable"

Output:

```
import com.rlux.test.elements.ExtJSSelectList;

private ExtJSSelectList categoryExtJSSelectList;

public ExtJSSelectList getCategoryExtJSSelectList() {
    return categoryExtJSSelectList;
}

public void setCategoryExtJSSelectList(ExtJSSelectList categoryExtJSSelectList) {
    this.categoryExtJSSelectList = categoryExtJSSelectList;
}
```

Any concerns for such functionality, and any feedback what to focus on if we would want to implement it?

mach6:
Closing this issue. #91 is tracking the outstanding documentation need for this enhancement

PR #90, opened 2015-04-13
Implements #81: Code-Generator: Add support for custom elements

renelux:
Implementation for feature request #81

### DIFF
--- a/codegen/src/main/java/com/paypal/selion/elements/HtmlSeLionElement.java
+++ b/codegen/src/main/java/com/paypal/selion/elements/HtmlSeLionElement.java
@@ -16,6 +16,7 @@
 package com.paypal.selion.elements;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import com.paypal.selion.plugins.GUIObjectDetails;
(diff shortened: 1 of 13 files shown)


codegen/src/main/java/com/paypal/selion/elements/HtmlSeLionElement.java
```diff
@@ -25,20 +26,20 @@
...
-    public static HtmlSeLionElement LINK = new HtmlSeLionElement("Link", true, true);
-    public static HtmlSeLionElement LABEL = new HtmlSeLionElement("Label", true, true);
-    public static HtmlSeLionElement IMAGE = new HtmlSeLionElement("Image", true, true);
-    public static HtmlSeLionElement FORM = new HtmlSeLionElement("Form", true, true);
-    public static HtmlSeLionElement DATE_PICKER = new HtmlSeLionElement("DatePicker", true, true);
-    public static HtmlSeLionElement CHECK_BOX = new HtmlSeLionElement("CheckBox", true, true);
-    public static HtmlSeLionElement CONTAINER = new HtmlSeLionElement("Container", true, false);
-    public static HtmlSeLionElement BASE_CLASS = new HtmlSeLionElement("baseClass", false, false);
-    public static HtmlSeLionElement PAGE_TITLE = new HtmlSeLionElement("pageTitle", false, false);
+    public static HtmlSeLionElement TEXT_FIELD = new HtmlSeLionElement("com.paypal.selion.platform.html", "TextField", true, true);
```
mach6: Minor suggestion: Make `com.paypal.selion.platform.html` a `public static final String` in this class

codegen/src/main/java/com/paypal/selion/elements/IOSSeLionElement.java
```diff
@@ -26,40 +27,40 @@
...
-    public static IOSSeLionElement UIATableView = new IOSSeLionElement("UIATableView", true);
-    public static IOSSeLionElement UIATarget = new IOSSeLionElement("UIATarget", true);
-    public static IOSSeLionElement UIATextField = new IOSSeLionElement("UIATextField", true);
-    public static IOSSeLionElement UIATextView = new IOSSeLionElement("UIATextView", true);
-    public static IOSSeLionElement UIAToolbar = new IOSSeLionElement("UIAToolbar", true);
-    public static IOSSeLionElement UIATouchScreen = new IOSSeLionElement("UIATouchScreen", true);
-    public static IOSSeLionElement UIAWebView = new IOSSeLionElement("UIAWebView", true);
-    public static IOSSeLionElement UIAWindow = new IOSSeLionElement("UIAWindow", true);
-    public static IOSSeLionElement BASE_CLASS = new IOSSeLionElement("baseClass", false);
+    public static IOSSeLionElement UIAButton = new IOSSeLionElement("org.uiautomation.ios.UIAModels", "UIAButton", true);
```
mach6: Same minor suggestion Re: `public static final String`

codegen/src/main/java/com/paypal/selion/elements/HtmlSeLionElement.java
renelux: Do you mean in the class AbstractSeLionElement?

```
private String element;
private boolean uiElement;
private String elementPackage;
```

This is how the class looks now so I kept it the same, you want me to update all to public static final?

mach6:
:+1: @renelux this look good to me but I'm biased since we worked most of the details out together over IRC.

I'm going to ask @sebady @krmahadevan or @fakrudeen78 to chime in.

codegen/src/main/java/com/paypal/selion/elements/HtmlSeLionElement.java
```diff
@@ -53,6 +56,20 @@
...
+     * Custom elements are inserted before SeLion elements, if you use the same name it will overwrite the existing element.
+     * 
+     * @param element string of the qualified class
+     */
+    public static void registerElement(String element) {
+        List<HtmlSeLionElement> temp = new ArrayList<HtmlSeLionElement>(Arrays.asList(values));
+        
+        temp.add(0, new HtmlSeLionElement(HtmlElementUtils.getPackage(element), HtmlElementUtils.getClass(element), true, false));
+        
+        values = temp.toArray(new HtmlSeLionElement[temp.size()]);
```
sebady, line 71: So the "overwrite" of an existing standard SeLion element value should at least emit a log message / warning here?  I assume this is not going to be a typical use case here with custom elements? Is there really a need to allow this use case since the order of the elements are matched is important and I can see this tripping up folks who are not careful.

renelux, line 71: I believe that the implementation has to be like this. For example in #81 I give an example of a ExtJS SelectList. This element is not compliant with <select> and the SeLion Element doesn't work with it. So I want to create a CustomElement similar to the SelectList if I don't put them in the beginning I could not use a name like ExtJSSelectList as the normal SelectList would be picked up first. So that would limit the users naming convention significantly. 

What about adding a check before elements get add if something with that name already exists? If it does I throw an exception. For the users that want to specifically overwrite, I can add a parameter to disable that check.

sebady, line 71: Can we just emit a log message/warning if the custom element will "over-write" a SeLion standard element?  That should be sufficient.

renelux, line 71: I have updated my code, in the CodeGeneratorMojo I check if the element is an existing SeLion element. If so it will emit a custom log message to make the user aware of this.

sebady:
Rene,  When I merged this I noticed it broke the BasicPageImplTest.  Can you look at what additional change is necessary to get BasicPageImplTest to run successfully?

renelux:
@sebady I just updated the BasicPageImpl. And the BasicPageImplTest is now passing.

As the container element is an AbstractElement during initialization it would identify it as a normal element. As the check for AbstractElement came before the check for Container. I changed the order so tt now first checks if the element is a Container, if not it will check if it is a AbstractElement

sebady:
merged

PR #92, opened 2015-04-14
Enable RuntimeReporter's Page Source support for mobile applications

ILikeToNguyen:
Main changes are refactoring WebLog/AppLog and WebReporter/MobileReporter to be merged together since there are no differences needed in code for web vs mobile.  getPageSource works for both.  Also removed 2 fields from WebLog/AppLog.  Moved PageContents and BaseLog(merged class) to internal package since end user shouldn't be dealing with these classes.


client/src/main/resources/templates/RuntimeReporter/index.html
```diff
@@ -22,8 +22,20 @@
         <link rel="stylesheet" href="../css/bootstrap-table.min.css">
         <!-- HTML5 shim and Respond.js IE8 support of HTML5 elements and media queries -->
         <!--[if lt IE 9]><script src="../bower_components/html5shiv/dist/html5shiv.js"></script><script src="../bower_components/respond/dest/respond.min.js"></script><![endif]-->
-        <style type="text/css">.modal-dialog {width: 95%; min-height: 95%; padding: 0;}.modal-content {min-height: 100%; }
-        </style>
+        <style type="text/css">.modal-dialog {width: 95%; min-height: 95%; padding: 0;}.modal-content {min-height: 100%; } </style>
```
ILikeToNguyen: Note this code is here temporary until I fix the next bug in this sprint.  ]
Issue is when clicking the page source button or image, no model dialog comes up because webpage complains the js functions aren't defined.
Main.js isn't being read in correctly which has these 2 js functions defined in it.  I will work on that and fix that in another PR.

client/src/main/java/com/paypal/selion/reports/runtime/Gatherer.java
```diff
@@ -41,7 +44,7 @@ static String saveGetLocation(WebDriver driver) {
                 location = driver.getCurrentUrl();
             }
         } catch (Exception exception) {
-            logger.log(Level.WARNING, "Current location couldn't be retrieved by getCurrentUrl().", exception);
+            logger.log(Level.FINEST, "Current location couldn't be retrieved by getCurrentUrl().", exception);
```
mach6: @ILikeToNguyen  Why the change to FINEST here?

client/src/main/java/com/paypal/selion/platform/grid/MobileNodeType.java
```diff
@@ -55,6 +57,12 @@ public static String getSupportedMobileNodesAsCSV() {
      * @return - A {@link MobileNodeType} enum that represents a SeLion compliant mobile node type.
      */
     public static MobileNodeType getMobileNodeType(String mobileNodeType) {
+        // this is a mandatory field ??? should we default it to appium if they haven't specified since appium supports
+        // both ios and android
+        if (StringUtils.isEmpty(mobileNodeType)) {
+
```
mach6: Why this empty `if` statement... We can address the mandatory-ness of `mobileNodeType` with a different change.

client/src/main/java/com/paypal/selion/internal/reports/model/BaseLog.java
```diff
@@ -13,47 +13,79 @@
...
 
-    protected AbstractLog() {
+    public BaseLog() {
     }
 
-    protected AbstractLog(String s) {
+    /**
+     * parse the string and build the object.
+     * 
+     * @param s
```
mach6: please complete the javadoc -- `s` is what?

client/src/main/java/com/paypal/selion/internal/reports/model/BaseLog.java
```diff
@@ -13,47 +13,79 @@
...
     protected String screen;
+    // the url of the current page.  This applies only to webtests and mobiletests that are running with safari, etc.
     protected String location;
+    // file location of the page source associated with this log message
+    protected String href;
 
     /**
      * Add custom parsing of a String that represents a line in the log.
      * 
      * @param part
```
mach6: please complete the javadoc -- `part` is what?

client/src/main/java/com/paypal/selion/platform/grid/MobileNodeType.java
ILikeToNguyen: Sloppy me.  We had a discussion where should we force the user to enter in a mobileNodeType(current implementation), or default it to appium for them if they dont explicitly define a mobileNodeType.  I was in thought process here and forgot to take this out.  I will take it out and address the mandatory-ness in a different change.

client/src/main/java/com/paypal/selion/internal/reports/model/BaseLog.java
```diff
@@ -13,47 +13,79 @@
...
+ * 
  */
-public abstract class AbstractLog {
+public class BaseLog {
 
     protected static SimpleLogger logger = SeLionLogger.getLogger();
 
-    protected String type;
     protected String msg;
+    // file location of the screenshot associated with this log message
```
mach6: since this is a `protected` field, please use javadoc comment instead of `//`

client/src/main/java/com/paypal/selion/internal/reports/model/BaseLog.java
```diff
@@ -149,4 +155,15 @@ public String getScreenURL() {
     public void setScreen(String screen) {
```
mach6, line 150: please add javadoc for this method

client/src/main/java/com/paypal/selion/internal/reports/model/BaseLog.java
```diff
@@ -149,4 +155,15 @@ public String getScreenURL() {
...
         this.screen = screen;
     }
+
+    public String getHref() {
```
mach6, line 177: please add javadoc for this method

client/src/main/java/com/paypal/selion/internal/reports/model/BaseLog.java
```diff
@@ -149,4 +155,15 @@ public String getScreenURL() {
...
+        if (href == null) {
+            href = "";
+        }
+        return href.replaceAll("\\\\", "/");
+    }
+
+    public void setHref(String href) {
```
mach6, line 189: please add javadoc for this method

client/src/main/java/com/paypal/selion/reports/runtime/Gatherer.java
ILikeToNguyen: currentLocation is valid for browser testing on mobileapps such as when the user is using safari app on appium-ios.  If a user is testing a nativeapp such as InternationalMountains.app, an exception will always be thrown on the call getCurrentURL.  I didnt want red text printed to the console for this scenario when testing native apps so i changed it to finest.  I didnt choose fine because i didnt want to clutter up the logs who is by default logging fine.

mach6: Okay, that makes sense.  I still feel we can do better :)

Though, I'm generally not in favor of `instanceof`, would doing an `instanceof` check on `driver` help in any way here? Sure, we won't be able to rely on `driver` with absolute certainty to tell us if it is a mobile `driver` AND we won't know anything about the `app` under test. Just a thought. Otherwise, my only suggestion for improvement is to change the logged message a bit to add something like `this can be safely ignored if testing a non-web mobile application`

client/src/main/resources/templates/RuntimeReporter/index.html
mach6: Let's leave this out of this PR and fix it for real later ..  Yes, I'm aware I won't be able to see the full impact of your improvement in this PR without this change... I can use the old-school SeLion Report though. :)

client/src/main/java/com/paypal/selion/internal/reports/model/package-info.java
```diff
@@ -0,0 +1,4 @@
+/**
+ * Classes related web/mobile logging, extracting HTML source and taking screen shots to assist in test reporting.
```
mach6: Please reword: suggest simplification: Something like: `Model classes to assist in test reporting.`

mach6:
@ILikeToNguyen This PR looks mostly okay to me. Please incorporate the feedback and update your branch.

client/src/main/java/com/paypal/selion/reports/runtime/SeLionReporter.java
```diff
@@ -37,10 +39,9 @@
     protected static DataSaver saver = null;
     protected String baseFileName = UUID.randomUUID().toString();
 
-    protected WebDriver driver;
```
ILikeToNguyen, line 40: FYI this change was not here when initial peer review was done.  Noticed it and decided that I would rather have this class not hold a reference to driver when we have static getter method already for this.

client/src/main/java/com/paypal/selion/reports/runtime/Gatherer.java
```diff
@@ -41,7 +44,10 @@ static String saveGetLocation(WebDriver driver) {
                 location = driver.getCurrentUrl();
             }
         } catch (Exception exception) {
-            logger.log(Level.WARNING, "Current location couldn't be retrieved by getCurrentUrl().", exception);
+            logger.log(
```
ILikeToNguyen, line 47: I decided to change the log from including the full exception stack which can eat up 20 lines of the console on every log call to just the exception name and message.

I'm still not 100% happy with the current way this is because, when we implement page object models for mobile testing just like we have Button, Link, etc for WebTest, when a user is running a flow on appium-ios for example, every app flow action such as click will cause a SeLionReporter.log call which will print a 4 line warning.  

Below is what will be added to the console with appium-android with this warning message.  4 lines.

1 thing to note: the only place i think that actually reads this 'location' value in BaseLog is from HtmlReporter, the old report.  So might be worth just logging it to finer since this is our old report.  Or perhaps we need to add this to our new Reporter since it was in our old report...

04.17.2015 00:35:08.072,[22] com.paypal.selion WARNING Current location couldn't be retrieved by getCurrentUrl(). This can be SAFELY IGNORED if testing a non-web mobile application. Reason: org.openqa.selenium.WebDriverException: Not yet implemented. Please help us: http://appium.io/get-involved.html (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 12 milliseconds
Build info: version: '2.45.0', revision: '5017cb8e7ca8e37638dc3091b2440b90a1d8686f', time: '2015-02-27 09:10:26'
System info: host: 'LM-AUN-00874810', ip: '127.0.0.1', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.9.5', java.version: '1.7.0_45'
Session ID: 8aed842a-6818-4f42-b915-bd56a2fa7c09
Driver info: com.paypal.selion.platform.grid.SeLionAppiumAndroidDriver
Capabilities [{platform=LINUX, app=/Users/antnguyen/dev/repo/SeLion/client/src/test/resources/apps/selendroid-test-app-0.15.0.apk, acceptSslCerts=true, javascriptEnabled=true, browserName=Android, mobileNodeType=appium, networkConnectionEnabled=true, unexpectedAlertBehaviour=ignore, desired={platformVersion=5.0.1, app=/Users/antnguyen/dev/repo/SeLion/client/src/test/resources/apps/selendroid-test-app-0.15.0.apk, javascriptEnabled=true, acceptSslCerts=true, platformName=Android, deviceName=Android Emulator, name=com.paypal.selion.android.sample.AppiumAndroidNativeAppTest:testWithNativeApp(), mobileNodeType=appium, unexpectedAlertBehaviour=ignore, takesScreenshot=true, version=}, locationContextEnabled=false, version=, platformVersion=5.0.1, databaseEnabled=false, platformName=Android, deviceName=Android Emulator, name=com.paypal.selion.android.sample.AppiumAndroidNativeAppTest:testWithNativeApp(), webStorageEnabled=false, warnings={}, takesScreenshot=true}]

mach6, line 47: @ILikeToNguyen 

>  the only place i think that actually reads this 'location' value in BaseLog is from HtmlReporter, the old report. So might be worth just logging it to finer since this is our old report. Or perhaps we need to add this to our new Reporter since it was in our old report...

I suspect we probably do need to add it to the new report.

That said, we still need to do something better here. This is the one part of this PR that is keeping me from merging.. I'm not super happy seeing this message on the console. Also, if the exception is caught for a non-appium case I now get this additional text which, to me is now noise (for the non-appium case) -- `This can be SAFELY IGNORED if testing a non-web mobile application` 

How about checking the exception message in the `catch` block for `Not yet implemented ... "appium.io"` and if it exists, log a different message like you are now, but maybe more concise.

``` java
logger.log(Level.WARNING, "Appium does not support getCurrentUrl(). This warning can be safely ignored.");
```

Otherwise, log the message as before with 

``` java
logger.log(Level.WARNING, "Current location couldn't be retrieved by getCurrentUrl().", exception);
```

@krmahadevan @fakrudeen78 @sebady do any of you have a better suggestion?

All - do any other drivers hit this `catch` block when `getCurrentUrl()` is called?

client/src/main/java/com/paypal/selion/internal/reports/model/BaseLog.java
```diff
@@ -13,47 +13,45 @@
...
+    protected String href;
 
-    /**
-     * Add custom parsing of a String that represents a line in the log.
-     * 
-     * @param part
-     */
-    abstract protected void parse(String part);
+    public BaseLog() {
+    }
```
mach6, line 46: why isn't there some default initialization for the member variables here?

ILikeToNguyen, line 46: The member variables are purposefully defaulting to null.  There is code checks for null which means that this log doesnt contain a screenshot or a pagesource.

client/src/main/java/com/paypal/selion/reports/runtime/Gatherer.java
ILikeToNguyen, line 47: "How about checking the exception message in the catch block for Not yet implemented ... "

I started going down that route but different drivers will throw a different message.  Id have to have a pretty ugly if else if or an lengthy || ||
Appium-IOS does : Not implemented in this context, try switching into or out of a web view
Appium-Android does: Not yet implemented. Please help us: http://appium.io/get-involved.html
IOS-driver:

client/src/main/java/com/paypal/selion/internal/reports/model/BaseLog.java
mach6, line 46: Thanks. I missed this.

client/src/main/java/com/paypal/selion/reports/runtime/Gatherer.java
krmahadevan, line 47: @mach6 
Here's my take on this...

We are returning a default value when we can't extract the currentUrl via the web driver. The warning is not going to do any value addition here  except for clog the console. So instead of worrying about the verbiage of the error message can we think of retaining the current error message and just change the level to `DEBUG` ? That way the console is not going to be clogged.. user is not misled and the curious ones can still find this info in the log files.

If we are determined to log this at the warning level but want to tweak the error message then I would suggest that we do something like below

```
((HasCapabilities) d).getCapabilities().getBrowserName(); // extract browser name
```

Enhance `com.paypal.selion.platform.grid.BrowserFlavors` (or create a new internal only class) to have a method which would basically accept the browser name as returned from the capability and then filter it out as `web` or `mobile` (We can have sub flavors as well) and then based on the enum it returns, we can display the corresponding error msg as well...

```
public enum ErrorType {
    APPIUM("appium", "Not supported in appium"),
    IOS("iosDriver", "Need help in implementing this in IOS Driver"),
    SELENDROID("selendroid", "This is currently being worked upon");
    private ErrorType(String type, String errorMsg) {
        this.type = type;
        this.errorMsg = errorMsg;
    }
    private String type;
    private String errorMsg;

    public static ErrorType getErrorMsgFor(String browserName) {
        if (IOSCapabilities.ipad().getBrowserName().equals(browserName)) {
            return IOS;
        }
        //rest of the code goes here
    }
}

```

So in the error msg section we would just do 

```
logger.log(Level.WARNING, ErrorType.getErrorMsgFor(((HasCapabilities) d).getCapabilities().getBrowserName()));
```

Would that work here ?

mach6, line 47: @krmahadevan Yes. I think that would help but I feel @ILikeToNguyen first instinct to log to `FINER` was the right approach. I'll make the change and merge this.

mach6:
@ILikeToNguyen I amended this PR and merged it.

PR #31, opened 2014-10-01
Fix an NPE when kicking off a Mobile test and update the IOSTest

azaytsev:

PR #94, opened 2015-04-15
Fixes #31: Document ElementEventListeners.

renelux:


html/capture-selion-events.html
```diff
@@ -0,0 +1,37 @@
+<div class="selion-content-div">
+  <h1 id="what-is-events" class="page-header">Working with SeLion events</h1>
```
mach6: I think we need to be specific throughout this document and not confuse the user.. `events` and `element events` are not necessarily the same thing.

html/capture-selion-events.html
```diff
@@ -0,0 +1,37 @@
...
+      </li>
+      <li>Provide the fully qualified class name of your listener via one of the following ways:</li>
+      <ul>
+        <li>The JVM argument <code>SELION_ELEMENENT_EVENT_LISTENER</code>
+          <pre><code class="language-sh">-DSELION_ELEMENT_EVENT_LISTENER=com.paypal.test.listeners.DebugScreenshotListener</code></pre></li>
+        <li>As a TestNG suite file parameter <code>elementEventListener</code>.
+          <pre data-src="../sample-files/events-sample2.xml"></pre>
+        </li>
+      </ul>
+      If you have more than one listener that you would want to plug into, then provide them as comma separated values.
```
mach6: change to `plug in` vs `plug into`

html/capture-selion-events.html
```diff
@@ -0,0 +1,37 @@
...
+
+  <p>
+    SeLion events are primarily those which are related to some action that you perform through elements provided by SeLion such as "click" (or) "select" etc.
+  </p>
+
+  <h2 id="getting-started">Getting Started with Event Handling in SeLion</h2>
+
+  <p>
+    SeLion offers support to hook into events that are generated by SeLion elements. These steps can be followed to start hooking into
+    events using SeLion. As an example we are going to create a DebugScreenshotListener which adds screenshots, that show data that was inputted to the page.
```
mach6: Some minor edits need to happen on this text;

`As an example we are going to create a DebugScreenshotListener which adds screenshots, that show data that was inputted to the page.`

Could be

`As an example we are going to create a <code>DebugScreenshotListener</code> which takes a screenshot and adds it to the report before any element is clicked on.`

sample-files/MyElementEventListener.java
```diff
@@ -0,0 +1,117 @@
...
+import com.paypal.selion.platform.html.support.events.Checkable;
+import com.paypal.selion.platform.html.support.events.Clickable;
+import com.paypal.selion.platform.html.support.events.Deselectable;
+import com.paypal.selion.platform.html.support.events.Selectable;
+import com.paypal.selion.platform.html.support.events.Submitable;
+import com.paypal.selion.platform.html.support.events.Typeable;
+import com.paypal.selion.platform.html.support.events.Uncheckable;
+import com.paypal.selion.reports.runtime.WebReporter;
+
+public class DebugScreenshotListener extends AbstractElementEventListener {
```
mach6, line 27: @renelux On the `gh-pages` branch we try to keep samples to 2 spaces per indent and 80 columns wide. This helps (a bit) with browsers/screen that do not have a wide viewport. Sorry for the deviation from the standard on the `develop` branch. It's the a low-tech solution, i know. Will you please update this sample source

renelux:
@mach6 I just updated the docs according to your suggestions.

mach6:
@renelux I amended your commit and merged it.. Thanks!

PR #96, opened 2015-04-20
Runtime Reporter: clicking on page source or screenshots broken

ILikeToNguyen:
Issue is that a modal dialog is suppose to come up when clicking those things.

Problem was that the javascript methods to perform the modal dialog was in the main function scope which isn't accessible globally by the html on click.  Moved the 2 java script methods out of main function into global scope.

Also made cursor turn into a hand when hovering over the screenshot image so user knows its clickable.


mach6:
@ILikeToNguyen I have made minor tweaks to your commit and have merged it with this commit - https://github.com/paypal/SeLion/commit/632ff1ed150600916e7d8f4fbaf2b948bbbeb17a

PR #98, opened 2015-04-23
Fix issue where page validator isVisible and isEnabled would always throw an exception

renelux:


mach6:
what was the exception prior to the change?

renelux:
ElementNotFoundException

We first checked if element is not found, then checked if it was visible which can't work.

mach6:
:+1: LGTM .. Will merge soon.

mach6:
this is merged.

PR #100, opened 2015-04-24
RuntimeReporter: add tooltip to display passed/fail information

ILikeToNguyen:
Added tool tip for the whole row in addition to just hovering over the colored numeric icon.  If a td object has a specified tool tip, that overrides the row's tooltip as intended.


mach6:
:+1: merged. thank you

PR #109, opened 2015-06-08
Test updates and misc re-factoring and/or bug fixes

mach6:
- Fixed bug in AbstractTestSession wher closeSession() would
  attempt to start a session when one was never started. Found
  by testing with an invalid browser type.
- Updated many "server" tests to use mocks for various needs
  -- such as protecting SELION_HOME_DIR, injecting some dummy
  process names for various tests, etc.
- Removed unnecessary code in ConfigParser.java, also changed
  method getInstance() to parse()
- Changed group to "functional" for previously classified
  "unit" tests in client which were requiring a browser session
- Relocated getExecutableNames() from ProcessNames.java to
  FileExtractor.java (the only consumer).
- Reduce visibility of many methods in server which did not
  need to be public
- Updated DummyArchive.zip and DummyBz2Archive.tar.bz2 to
  NOT contain executables named "phantomjs"
- Fix issue where JarSpawner would process CLASSPATH icorrectly
  on Windows


mach6:
@krmahadevan @sebady @fakrudeen78  Can I please have at least one of you review this? Thanks!

server/src/test/java/com/paypal/selion/utils/AuthenticationHelperTest.java
```diff
@@ -15,26 +15,34 @@
...
-        SeLionGridConstants.SELION_HOME_DIR = SystemUtils.USER_DIR + "/src/test/resources/";
+    public void mockTestPath() throws Exception {
+        mockStatic(SeLionGridConstants.class);
+        Whitebox.setInternalState(SeLionGridConstants.class, "SELION_HOME_DIR", SystemUtils.USER_DIR + "/target/temp/");
+        new File(SeLionGridConstants.SELION_HOME_DIR).mkdirs();
     }
 
     @Test
     public void testAuthentication() {
+        System.out.println(SeLionGridConstants.SELION_HOME_DIR);
```
mach6: This is an oversight. I'll get rid of it before I merge.

client/src/main/java/com/paypal/selion/platform/grid/AbstractTestSession.java
```diff
@@ -198,13 +198,13 @@ public final String getTestName() {
      */
     public final void closeSession() {
         logger.entering();
-        new SauceLabsHelper().embedSauceLabsJobUrlToTestReport();
-        if (Grid.driver() != null) {
+        if (this.isStarted && Grid.driver() != null) {
```
sebady: Why not use the method isStarted()?

sebady:
The rest of these changes look ready for merge

PR #114, opened 2015-06-30
changes to run locally and misc fixes and/or refactoring

mach6:
1. Client now uses selion enhanced grid for local runs
2. ios-driver 0.6.6-snapshot is gone but optional 
3. Client has no selenium-sever dependencies unless running locally. Server dependencies for local runs are downloaded on demand and/or configured in by user
4. Misc other changes ... view commit message(s) for additional details


server/src/main/java/com/paypal/selion/grid/RunnableLauncher.java
```diff
@@ -0,0 +1,61 @@
...
+package com.paypal.selion.grid;
+
+/**
+ * Interface which extends {@link Runnable} and adds some additional methods
+ */
+public interface RunnableLauncher extends Runnable {
+    /**
+     * Shutdown the launcher
+     */
+    abstract void shutdown();
```
krmahadevan: `abstract` keywords are redundant in an interface. Please have them removed.

mach6:
@fakrudeen78 @sebady @krmahadevan submitted this for initial feedback. The one item I'm aware of that needs additional improvement is shutdown() logic for the process based RunnableLauncher's

server/src/main/java/com/paypal/selion/grid/RunnableLauncher.java
mach6: Yes. Sure are. Will fix

server/src/main/java/com/paypal/selion/grid/AbstractBaseLauncher.java
```diff
@@ -0,0 +1,408 @@
...
+
+    /**
+     * Get program arguments to pass
+     *
+     * @return the program arguments to pass represented as an array of {@link String}
+     * @throws IOException
+     */
+    String[] getProgramArguments() throws IOException {
+        LOGGER.entering();
+        List<String> args = new LinkedList<String>();
```
krmahadevan: Do we need to iterate through every element for this initialization ?

Can we do something like below ? 

`List<String> args = new LinkedList<String>(commands);`

server/src/main/java/com/paypal/selion/grid/MobileProcessLauncher.java
```diff
@@ -0,0 +1,121 @@
...
+     */
+    @Override
+    String[] getProgramArguments() throws IOException {
+        LOGGER.entering();
+        List<String> args = new LinkedList<String>();
+        args.addAll(Arrays.asList(super.getProgramArguments()));
+
+        // add the defaults which we don't already have a value for
+        Iterator<Entry<String, JsonElement>> iterator = defaultArgs.entrySet().iterator();
+        while (iterator.hasNext()) {
```
krmahadevan: Since we are interacting with every element in this Set, why not just use 

```
for (Map.Entry<String, JsonElement> eachArg : defaultArgs.entrySet()) {
...
}
```

server/src/main/java/com/paypal/selion/grid/MobileProcessLauncher.java
```diff
@@ -0,0 +1,121 @@
...
+            }
+            args.remove(filter);
+        }
+
+        LOGGER.exiting(args.toString());
+        return args.toArray(new String[args.size()]);
+    }
+
+    void printUsageInfo() {
+        StringBuffer usage = new StringBuffer();
```
krmahadevan: Should we resort to just using `StringBuilder` ?

server/src/main/java/com/paypal/selion/grid/ProcessLauncherOptions.java
```diff
@@ -0,0 +1,160 @@
...
+    private boolean includeWebDriverBinaryPaths = true;
+    private boolean continuousRestart = true;
+    private boolean fileDownloadCleanupOnInvocation = true;
+    private boolean fileDownloadCheckTimeStampOnInvocation = true;
+
+    /**
+     * Enable/Disable clean up of previously downloaded artifacts for subsequent calls to {@link FileDownloader} .
+     * Default: enabled.
+     */
+    public ProcessLauncherOptions setFileDownloadCleanupOnInvocation(boolean val) {
```
krmahadevan: Since the method intends to `enabled/disable`, should we rename this method to `toggleFileDownloadCleanupOnInvocation` ?

Similarly for the other methods as well ?

server/src/main/java/com/paypal/selion/grid/AbstractBaseProcessLauncher.java
```diff
@@ -0,0 +1,359 @@
...
+
+        LOGGER.exiting();
+    }
+
+    /**
+     * Start a process based on the commands provided.
+     *
+     * @param squelch
+     *            Whether to show command executed as a logger.info message
+     * @param cmdLine
```
krmahadevan: Please fix the javadocs.`cmdLine` is no longer a valid argument.

server/src/main/java/com/paypal/selion/grid/JarSpawner.java
```diff
@@ -116,317 +84,83 @@ private void printUsageInfo() {
...
+        CloseableHttpClient client = HttpClientBuilder.create().build();
+        StringBuilder url = new StringBuilder();
         try {
-            FileUtils.forceMkdir(new File(DOWNLOADS_DIR));
-            FileUtils.forceMkdir(new File(LOGS_DIR));
-            copyFileFromResources(DOWNLOAD_JSON_FILE_RESOURCE, DOWNLOAD_JSON_FILE);
-            copyFileFromResources(SELION_CONFIG_FILE_RESOURCE, SELION_CONFIG_FILE);
-            ConfigParser.setConfigFile(SELION_CONFIG_FILE);
+            url.append("http://");
+            url.append(getHostPortAndServletPath("NodeForceRestartServlet"));
```
krmahadevan: Please use `url.append(getHostPortAndServletPath(NodeForceRestartServlet.class.getSimpleName()));`

krmahadevan:
Minor observation

We can remove `implements LocalServerComponent` from the below classes 
- LocalHub
- LocalSelendroidNode
- LocalNode
- LocalIOSNode

since all of the above 4 classes extend from `AbstractBaseLocalServerComponent` and since the base class already implements `LocalServerComponent`

On a related observation, I believe some of the discussions we were having were related to getting rid of the need for `LocalHub` and `LocalNode` for browser based automation and instead directly start leveraging the existing concrete `WebDriver` implementations. Are those going to be a separate change ?

Also I believe that for starting a test locally against iOS driver (or) selendroid, we can very well just spawn the LocalIOSNode/ LocalSelendroidNode and then work with creating a RemoteWebDriver instance using it. If that is the route that we take then I believe we can very well get rid of the LocalHub and LocalNode classes for sure.

server/src/main/java/com/paypal/selion/pojos/SeLionGridConstants.java
```diff
@@ -235,4 +197,9 @@ static String adjustFileSeparatorForPlatform(String path) {
      * Dash argument for displaying the help text
      */
     public static final String HELP_ARG = "-help";
+
+    /**
+     * Dash argument for disabling continuous restart nature of {@link JarSpawner}
+     */
+    public static final String SELION_NOCONTINUOS_ARG = "-noContinuousRestart";
```
krmahadevan: Please fix the typo

client/src/main/java/com/paypal/selion/platform/grid/LocalSelendroidNode.java
```diff
@@ -15,191 +15,168 @@
...
-
-        // Specify the aut folder so its contents will be monitored and read by Selendroid.
-        String autFolder = Config.getConfigProperty(ConfigProperty.MOBILE_APP_FOLDER);
-        if ((autFolder != null) && (!autFolder.trim().isEmpty())) {
-            args.add("-folder");
-            args.add(autFolder);
+    private void validateConfiguredOptions() {
+        // Make sure the configured internal selendroid server port is not already in use
+        int selendroidServerPort = Config.getIntConfigProperty(ConfigProperty.SELENDROID_SERVER_PORT);
+        if (PortProber.pollPort(selendroidServerPort)) {
```
krmahadevan: Either the `pollPort` method is incorrectly returning a value (or) we are missing a `!` condition here. Please have this fixed. Currently the implementation is such that `pollPort` returns a `true` if we can establish a socket bind on a given port. That would mean that the port is NOT already in use and its free to be used.

client/src/main/java/com/paypal/selion/platform/grid/LocalSelendroidNode.java
```diff
@@ -15,191 +15,168 @@
...
+            throw new IllegalArgumentException("Specified SELENDROID_SERVER_PORT is already in use.");
```
krmahadevan: This error message needs to be enhanced such that it calls out the actual port number in which we attempted to make a connection and that failed. This would give the end user to figure out what in their machine is already using that port and fix things accordingly. Else it becomes very difficult.

client/src/main/java/com/paypal/selion/platform/grid/LocalSelendroidNode.java
```diff
@@ -15,191 +15,168 @@
...
- * 
+ * A singleton that is responsible for encapsulating all the logic w.r.t starting/shutting down a local selendroid node.
  */
-class LocalSelendroidNode extends AbstractNode implements LocalServerComponent {
+@Beta
+final class LocalSelendroidNode extends AbstractBaseLocalServerComponent implements LocalServerComponent {
+    private static final SimpleLogger LOGGER = SeLionLogger.getLogger();
+    private static volatile LocalSelendroidNode instance;
+
+    static synchronized final LocalServerComponent getSingleton() {
```
krmahadevan, line 40: I don't think this method is being used anywhere.  Please remove this off.
Same case with the other classes as well.

krmahadevan:
I am having problems in getting the suite file `suites/SelendroidMobileSuite.xml` to run locally with this implementation. The root cause of the problem is that the implementation causes the download.json to be extracted out to my `~/.selion` folder and then `com.paypal.selion.grid.FileDownloader#checkForDownloads(com.paypal.selion.grid.RunnableLauncher.InstanceType, boolean, boolean)` downloads the `selenium-server-standalone-2.45.0.jar` jar. But then later when the implementation attempts at booting up the `com.paypal.selion.platform.grid.LocalSelendroidNode` node, the code flow eventually lands up in 
`com.paypal.selion.platform.grid.LocalSelendroidNode#boot` > `com.paypal.selion.platform.grid.AbstractBaseLocalServerComponent#boot` > `com.paypal.selion.grid.SelendroidJarSpawner#run` > `com.paypal.selion.grid.AbstractBaseProcessLauncher#run` > `com.paypal.selion.grid.FileDownloader#checkForDownloads(com.paypal.selion.grid.RunnableLauncher.InstanceType)` > `com.paypal.selion.grid.FileDownloader#checkForDownloads(com.paypal.selion.grid.RunnableLauncher.InstanceType, boolean, boolean)` wherein the below logic is preventing the selendroid jar from being downloaded 
```
        if ((lastModifiedTime == downloadFile.lastModified()) && checkTimeStamp) {
            return;
        }

```

Thus causing the selendroid node to never boot up.

This scenario can be recreated if you attempt at running the code freshly using the above mentioned suite file ( my `~/.selion` folder is empty and the jars are being downloaded ONLY now for the first time)

krmahadevan:
The current implementation causes problems for the very first time when someone wants to basically run a web test locally. This scenario becomes valid especially when the user is on a very slow network and the downloading is slow.

In my case `phantomjs-1.9.8-macosx.zip` took a long time to download and even before it could finish, I ended up triggering a 

```
Caused by: java.lang.IllegalStateException: Timed out waiting for LocalNode to initialize.
    at com.paypal.selion.platform.grid.AbstractBaseLocalServerComponent.waitForInitialization(AbstractBaseLocalServerComponent.java:113)
    at com.paypal.selion.platform.grid.AbstractBaseLocalServerComponent.boot(AbstractBaseLocalServerComponent.java:63)
    ... 31 more
```

krmahadevan:
@mach6 

I feel that we may perhaps have complicated things a little bit here.
I was hoping that we could be doing just the following to get this done at the local executions level.
- Introduce 1 new Configuration property that would tell SeLion how to spin off a new server. This would be the command that we would be using to spin off either a Selendroid node (or) an Appium node (or) an iOS Driver node.
- Get rid of LocalHub, LocalNode etc., and just have 1 class named "LocalRemoteNode" (or whatever we would like to call it as ) which would spin off a child process by using the command that the user has specified and target `@MobileTest` annotated test runs against it.
- For `@WebTest` annotated test methods which are to be running locally, we would be falling back to using `FirefoxDriver`, `InternetExplorerDriver`, `ChromeDriver` etc., and thus do away with the need for using a programmatically spun `Hub` and `Node`.

That way we don't need to bring in `Selion Grid` as a dependency on the `Client` and leave things as they are.

server/src/main/java/com/paypal/selion/grid/MobileProcessLauncher.java
mach6: Will change this

server/src/main/java/com/paypal/selion/grid/ProcessLauncherOptions.java
mach6: Good suggestion. Will change this

server/src/main/java/com/paypal/selion/grid/AbstractBaseProcessLauncher.java
mach6: Will fix

server/src/main/java/com/paypal/selion/grid/JarSpawner.java
mach6: Usage of `NodeForceRestartServlet.class.getSimpleName()` requires `HttpServlet` (sub-class of `NodeForceRestartServlet`) to be added as a `compile` dependency and included into the Selion-Grid (standalone) jar -- which I'm avoiding here. That said, `shutdown()` in all process based `RunnableLaunchers` is the area I called out as needs improvement.. Relying on `NodeForceRestartServlet` here as I have coded, is not a reliable method and one that does not work for `STANDALONE` and `HUB` ...

server/src/main/java/com/paypal/selion/pojos/SeLionGridConstants.java
mach6: Took me a min to see it, but yes, will fix it

client/src/main/java/com/paypal/selion/platform/grid/LocalSelendroidNode.java
mach6: This is a bug.. will fix

client/src/main/java/com/paypal/selion/platform/grid/LocalSelendroidNode.java
mach6: How about letting selendroid throw the error `can't bind to port` instead and just remove `validateConfiguredOptions` and the pre-existing `checkAndValidateParameters` methods outright?

client/src/main/java/com/paypal/selion/platform/grid/LocalSelendroidNode.java
mach6, line 40: It exists for two reasons;
- So a `@Test` method, which relys on `LocalGridManager` to boot the `LocalSelendroidNode` instance, can access the booted `LocalSelendroidNode` instance
- To override the implementation of `getSingleton()` in `AbstractBaseLocalServerComponent` which throws an `IllegalStateException` (by design).

mach6:
> I am having problems in getting the suite file suites/SelendroidMobileSuite.xml .....

This is a bug.  I introduced the ability to bypass the timestamp check in `FileDownloader` by setting `ProcessLauncherOptions` .. Although the option `setFileDownloadCheckTimeStampOnInvocation(false)` is set in `LocalSelendroidNode`, it seems to not be happening.. So, I'll look into this..

mach6:
> Caused by: java.lang.IllegalStateException: Timed out waiting for LocalNode to initialize.

Letting initialization run indefinitely did not feel like the right thing to do. Thus, the initialization timeout is currently configurable via `ConfigProperty.EXECUTION_TIMEOUT`.  Perhaps a a new config option would be more suitable. Another thought I had was to package these artifacts (the one that we would otherwise download via `FileDownloader`) in the SeLion jar and extract them (instead of downloading them) as long as the filename and checksum matches the content of the `download.json` file.

client/src/main/java/com/paypal/selion/platform/grid/LocalSelendroidNode.java
krmahadevan, line 40: Why do we need an explicit method such as this one to check if the node/hub is up and running ? In any case all we are doing is just spinning off a sub process and there is no assurance that a node was really up and running due to us spinning it off as a sub process until and unless we can actually access that standalone JVM and interact with it. So what are we asserting here ?
Shouldn't we just be resorting to invoking the Grid API calls i.e.,
- `grid/api/hub/` to get the Hub status
- `/grid/api/proxy/` to get the node status
- `/wd/hub/status` to get the Selendroid node status.

Once we take this route then we don't need a `getSingleton()` implementation in the base class nor the child classes.

client/src/main/java/com/paypal/selion/platform/grid/LocalSelendroidNode.java
krmahadevan: We could do that as well. But I have seen a lot of people not being able to understand the `cant't bind to port` error message that Java throws. So it would be a good value addition if we provide a good error message.

server/src/main/java/com/paypal/selion/grid/MobileProcessLauncher.java
krmahadevan: @mach6 
I saw the same StringBuffer being used in a couple of other places as well. So can we just take care of all of that in one shot ? I forgot to call that out before.

krmahadevan:
@mach6 
Here's the problem that I have with this :

Imagine the below scenario. Our user is working on an offline basis wherein he/she doesn't have an internet connection.
Now the user wants to basically run some functional tests against a locally running Tomcat server which hosts his/her web application for which the user wants to write automation tests.

Our current model will never let a user run his tests because it expects to download chromedriver/phantomJS driver/InternetExplorer driver etc., from the internet apart from downloading the selenium jars before he/she can run anything.

What if the user is behind a corporate proxy/firewall setup which prevents downloads from the internet ?

This to me is a big step in the reverse direction.

IMO we should delegate the responsibility of setting up the local environment to the user (even Selenium expects the users to do this as a pre-requisite and doesn't automagically do it for them) and just run against the setup environment.

Also if a user has already setup his environment (such as installing chromedriver and phantomJS driver via tools such as `brew` ) shouldn't we be honoring that setup rather than us try to setup things from the scratch ?

So for local runs we should just keep things simple by asking the user to provide us with the command to spin off a node and use that command to spin off a single sub process (ONLY for selendroid, iOS-driver and appium local runs use cases) and then have the tests run against that newly spun off sub process.

For web tests involving local runs we should use `ChromeDriver`, `InternetExplorerDriver`, `FirefoxDriver` etc., and do away with the local hub and local node.

client/src/main/java/com/paypal/selion/platform/grid/LocalSelendroidNode.java
mach6, line 40: `isRunning()` which makes an api call is a method which is accessed via the `RunnableLauncher`.. which can be accessed via `getLauncher()` on the `LocalServerComponent`.. hence the reason `getSingleton()` exists for testing.

mach6:
@krmahadevan I fail to see how downloading dependencies (in the case of brew or other) which requires an internet connection and pre-work is any different -- a user can pre-establish `~/.selion`, if needed. The means to specify alternative chromedriver, phantomjs, etc installations exists in the same fashion as it did prior to this change (we can and likely should do the same for ios-driver standalone and selenium-sever standalone). The only difference is by default, we are trying to setup the environment IFF those binary/artifact locations were not provided to us. The exception being selenium-server -- which MUST be included in the users pom (therefore already in the local maven repo) and should NOT (if it is, this was not the intent, and I need to fix it) be downloaded for a `LocalHub` or`LocalNode` when these are created via `ThreadedLauncher`

> For web tests involving local runs we should use ChromeDriver, InternetExplorerDriver, FirefoxDriver etc., and do away with the local hub and local node... 

This change, if we decide to go this route, can come later. My intent here was to address the multiple selenium-server version conflicts, ios-driver snapshot, multiple implementations (both in client and server) for spinning up `Sel` nodes/hubs etc and to keep the general approach of using `RemoteWebDriver` and a single access port (4444) everywhere in tact for now. In this effort we gained some things -- but, yes it does mean a couple new assumptions.

The changes I made to the "server" jar needed to happen regardless. We needed to support launching more than just a regular `Sel` node, hub, or standalone

krmahadevan:
@mach6 
I believe you are taking my example in the literal sense. This implementation assumes that the artifacts will be available in a pre-defined location. If its not available in that location it would be downloaded.

The point I am trying to make here is that, this assumption is an overhead due to the following reasons 
- How does the user know about this assumption ? Yes documentation could be one of the things. But then again not all things can be taken care of in that manner. For e.g., a user would perhaps have already "pre-setup" his machine by making the chromedriver, phantomJs binaries in predefined locations of his/her choice. Why are we expecting that they be provided again in a different location ?
- If the machine was already pre-setup by the user in terms of making the jars available and the binaries being available I am of the opinion that we should just use them.

With respect to using other web drivers, I believe that it should be done in this itself because doing that is going to get rid of the need for a local Hub and local node completely. That means that there is no need for the changes done at the server jar to be made.  The current implementation that we had (ie., a customized Grid with a plain vanilla node) itself should suffice. Launching more than just the regular node/hub and standalone is IMO not required because there is not much we are doing in the other standalones (such as appium, iOS driver, selendroid) apart from providing mechanisms for recycling a node via extra servlets.

sebady:
I see some value in having the SeLion client use our local hub and a local node.  This provides the ability to have an execution environment that is "closer" to what you will get from our Grid.   We can also support a direct runlocally mode where the browser driver instances are leveraged directly (from a user configurable location if necessary).  This is really a new capability though and I would recommend we try to support both modes of "runlocally".

server/src/main/java/com/paypal/selion/grid/AbstractBaseLauncher.java
mach6: will fix

server/src/main/java/com/paypal/selion/grid/MobileProcessLauncher.java
mach6: yep. makes sense. will change

mach6:
> This implementation assumes that the artifacts will be available in a pre-defined location. If its not available in that location it would be downloaded.

@krmahadevan this behavior is fixed in the most recent updates / commit.  Driver binaries are downloaded to and used in `SELION_HOME_DIR` as a last resort.  --- selenium is not downloaded anymore for LocalHub/LocalNode -- Also addressed your other misc feedback. Last, as we discussed offline (and for the reasons stated) this implementation continues to spin up a local hub and local node for `@WebTest` needs.

mach6:
@fakrudeen78 @sebady @krmahadevan @shelar Ping.. New commits on this PR. Feedback welcome. Note: I still have pending (not yet complete) changes for the one area I see outstanding with this PR -- improving the `AbstractBaseProcessLauncher#shutdown()` logic

krmahadevan:
@mach6 
Can you please help elaborate as to what is going to be done with #115 ?

Without us making a decision on that, I don't think it makes sense for us to continue vetting out this delivery.

I am stil not convinced that we need such an elaborate implementation to support local execution.

sundaramrajendran:
@mach6 

I think we can remove the capability of starting the Grid locally in SeLion client which will simplify the code:
- Can remove code for starting the grid
- Get rid of browser driver config and SELENIUM_RUN_LOCALLY variables 
- No need to download the Selenium-Server and Selion-Grid dependencies in client

Your approach simplify the user experience, but I think right now setting-up/starting our SeLion grid is very simple. So user can easily do it.

Just my thought on this...

mach6:
W.R.T to running local for web, etc. My intention and approach in this PR keeps the existing capabilities intact as much as possible. 

We can/should pursue other options after M5 and 1.0.0.

client/src/main/java/com/paypal/selion/platform/grid/LocalNode.java
```diff
@@ -15,141 +15,140 @@
...
+import org.openqa.selenium.net.NetworkUtils;
+import org.openqa.selenium.net.PortProber;
+import org.openqa.selenium.os.CommandLine;
+
+import com.paypal.selion.SeLionConstants;
 import com.paypal.selion.configuration.Config;
 import com.paypal.selion.configuration.Config.ConfigProperty;
+import com.paypal.selion.grid.LauncherOptions;
+import com.paypal.selion.grid.ThreadedLauncher;
+import com.paypal.selion.grid.LauncherOptions.LauncherOptionsImpl;
```
sebady: This file LauncherOptions seems to be missing from commit.

PR #121, opened 2015-07-15
Refactor data providers

sebady:
This pull request builds on changes in the dsimmons-changes branch to refactor our Data Providers into a standalone module.

This provides the benefit of allowing users to leverage our DataProviders with testng automation.


dataproviders/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+                </includes>
+            </resource>
+        </resources>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <suiteXmlFiles>
+                        <suiteXmlFile>${suiteXmlFile}</suiteXmlFile>
```
mach6: Make this `dataproviderSuiteXmlFile`... if you build from the parent directory with -DsuiteXmlFile=PhantomJS-Suite.xml this module will not build because there is no such suite.. This is why travis ci has failed your PR -- https://travis-ci.org/paypal/SeLion/builds/71008236

dataproviders/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+        <suiteXmlFile>src/test/resources/suites/Default-Suite.xml</suiteXmlFile>
+    </properties>
+
+    <build>
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
+                <filtering>true</filtering>
+                <includes>
+                    <include>selionbuildinfo.properties</include>
```
mach6: I don't see this as needed in this module. the class that consumes it is in `SeLion-Common`

dataproviders/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>commons-lang</groupId>
+            <artifactId>commons-lang</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
```
mach6, line 67: Do we absolutely need `lang3` AND `lang`? I've been trying to get rid of the dual dependencies here. You will see I changed all usage of `lang3` in `SeLion-Grid` for example

dataproviders/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>xml-apis</groupId>
+            <artifactId>xml-apis</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>dom4j</groupId>
+            <artifactId>dom4j</artifactId>
```
mach6, line 117: Can any dependencies be removed from `client` as a result of the refactoring in this PR? I don't think `dom4j` is used anywhere else, for example.

project-bom/pom.xml
```diff
@@ -100,6 +106,13 @@
                 <version>1.9.2</version>
             </dependency>
 
+            <dependency>
+                <artifactId>guava</artifactId>
+                <groupId>com.google.guava</groupId>
+                <!-- TODO check this version -->
+                <version>18.0</version>
```
mach6, line 106: 18.0 is the version Selenium 2.45.0 requires.  So, I think you can remove this `TODO`

dataproviders/src/test/resources/suites/Default-Suite.xml
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd">
+<suite thread-count="10" verbose="2" name="Dataprovider Test Suite" skipfailedinvocationcounts="false" junit="false"
+    parallel="methods" data-provider-thread-count="1" annotations="JDK">
+
+    <parameter name="browser" value="*chrome" />
```
mach6: shouldn't need this `parameter` for this suite.

common/pom.xml
```diff
@@ -74,6 +80,21 @@
         </dependency>
 
         <dependency>
+            <groupId>commons-logging</groupId>
+            <artifactId>commons-logging</artifactId>
+        </dependency>
```
mach6, line 85: similar question here... can we remove any dependencies in client, now that we have push some code into common?

dataproviders/pom.xml
sebady: I will fix this.

dataproviders/pom.xml
sebady, line 67: I left all the java imports as is.  I can update all the imports here to use lang3 instead and start getting off of lang.

dataproviders/pom.xml
sebady, line 117: I'll see if I can remove dom4j from client pom.

dataproviders/pom.xml
sebady, line 67: Actually just dropping "commons-lang" was fine.  We can use lang3 everywhere.

dataproviders/pom.xml
sebady, line 117: Actually there is no dom4j dependency in client pom.

common/pom.xml
sebady, line 85: client/pom.xml does not have a dependency on commons-logging

krmahadevan:
@sebady 
I have a few questions on this delivery.
I don't think we have yet come to a consensus on #114 Vs #115 .
Why is this PR targeting `dsimmons-changes` branch (which basically represents #114 ) rather than the `develop` branch ?

So I would recommend that you please hold off on this till we come to a consensus on that, (or) branch off of `develop` and raise a new Pull Request.

sebady:
This PR builds on dsimmons-changes branch only because it requires the common module introduction on that branch.   Having a new common module already in place make this dataprovider refactoring much cleaner to achieve at this point.

krmahadevan:
@sebady 
Then I believe we would need to extract out that part as a separate PR, vet that out, get it merged and have your branch rebase from `develop`. Else your PR is also likely to be stalled.

mach6:
114 is not stalled. Rather, it was not ready.

I was pretty clear in my comments in 115 that it should come as an iterative change after 114. As a result, we are targeting 114 first (when it's ready)..  I'm not going to piece-mill 114 into multiple PR's. All of the changes are relevant to the forward progress we are tracking in M5 (refactoring, code and dependency cleanup, and minimal feature changes). Furthermore 114 puts us in a place to do a pre-release (beta or RC) without changing the existing approach. 

We need to instead be talking about when and how the proposed changes in 115 come to light. My suggestion was after 1.0.0 but I could also see it before 1.0.0. I could also see a couple 1.0.0-beta releases between now and 1.0.0.. The point is we are close to some kind of release.. Generally speaking, outside of refactoring and optimizing our existing capabilities, we don't want serious changes between now and then.. We also don't want serious changes between now and  the 1.0.0.  Exceptions, for the right reasons and depending on timing/external considerations, etc are certainly acceptable. I feel 115 and this PR (121) are reasonable examples.. The important thing here is to figure out the timing and remain vigilant in our efforts to get to the 1.0.0 release.

krmahadevan:
@mach6 
I dont understand why #115 needs to come in as an iterative change over #114 (especially when it is eventually going to end up removing off a bunch of classes that were introduced in #114) 

> > Furthermore 114 puts us in a place to do a pre-release (beta or RC) without changing the existing approach.

I believe this is the same case with #115 as well. I am yet to hear of any reasons which is going to state that this isn't the case. I believe both @fakrudeen78 and @sundaramrajendran took a look at it and none of them seemed to have called out anything which would be of cause for concern.

> > Generally speaking, outside of refactoring and optimizing our existing capabilities, we don't want serious changes between now and then.. 

I am not sure if you took a closer look at #115, but it is also just refactoring and optimizing existing capabilities. I am not sure what are you calling out here as `serious` changes.

mach6:
>  (especially when it is eventually going to end up removing off a bunch of classes that were introduced in #114)

Let's see. In my opinion, 115 will remove classes which were pre-existing and possibly refactored as part of 114. Having capabilities in "selion-grid" which add value, whether or not they are used by "client", to me, is reasonable.

> I believe this is the same case with #115 as well. I am yet to hear of any reasons which is going to state that this isn't the case. 

I don't disagree. There are other considerations in 114 which are done that is not part of 115. dependency convergence, misc fixes, minor improvement, common, etc. Regardless of the way client spins up a  local web-only webdriver hub+node OR standalone+[chromedriver,phantomjs,etc] OR [chromedriver,phantoms,etc] in isolation (which imo, is the way we should do things for local web), the  other changes should be incorporated.

>  I am not sure if you took a closer look at #115, but it is also just refactoring and optimizing existing capabilities. I am not sure what are you calling out here as `serious` changes.

Yes, I have looked at it in detail and have considered incorporating the salient parts into 114. I'd change some of the approach but don't really consider the root of the proposed change to be `serious`.  I was speaking generically in my last comment about any potential change between now and release.

PR #124, opened 2015-07-21
refactor data providers into standalone module

sebady:
This pull request refactors our Data Provider functionality into a standalone module that allows clients to leverage our DataProviders with testng automation.


mach6:
@sebady I'm making changes and moving this to a branch.

mach6:
closing and re-opening against new branch.

client/src/test/resources/AssociativeArrayOfLists.yaml
```diff
@@ -1,6 +0,0 @@
-# Block associated array of lists (showing examples of both inline and blocked lists)
```
mach6, line 1: @sebady Is this an inadvertent removal? I see test skips in the new dataprovider module as a result of this file (and the other 6 below) removed as test resources

sebady, line 1: Yes this removal was accidental.  The files should have been moved into dataprovider module.

PR #131, opened 2015-07-24
Fixed critical and major violations highlighted out of sonar report

ysecaran:
Travis build: https://travis-ci.org/ysecaran/SeLion/builds/72437362 
@mach6 @fakrudeen78 the sonar violations are fixed and the travis build is green.
Please review and merge the same .

### DIFF
--- a/client/src/main/java/com/paypal/selion/platform/grid/MobileTestSession.java
+++ b/client/src/main/java/com/paypal/selion/platform/grid/MobileTestSession.java
@@ -49,8 +49,8 @@ public class MobileTestSession extends AbstractTestSession {
     private WebDriverPlatform platform;
     private MobileNodeType mobileNodeType;
 
-    private static String SAUCE_URL = "sauce-storage:";
-    private static String SELION_HUB_STORAGE = "selion-hub-storage";
+    private static final String SAUCE_URL = "sauce-storage:";
+    private static final String SELION_HUB_STORAGE = "selion-hub-storage";
 
     MobileTestSession() {
         super();
@@ -131,7 +131,6 @@ public class MobileTestSession extends AbstractTestSession {
         if (StringUtils.isNotBlank(getLocalConfigProperty(ConfigProperty.MOBILE_NODE_TYPE))) {
             mobileNode = getLocalConfigProperty(ConfigProperty.MOBILE_NODE_TYPE);
         }
-
         // Override values when supplied via the annotation
         if (deviceTestAnnotation != null) {
             if (StringUtils.isNotBlank(deviceTestAnnotation.appName())) {
@@ -181,6 +180,8 @@ public class MobileTestSession extends AbstractTestSession {
                 mobileNode = deviceTestAnnotation.mobileNodeType();
             }
             this.mobileNodeType = MobileNodeType.getMobileNodeType(mobileNode);
+            
+            initializeAdditionalCapabilities(deviceTestAnnotation.additionalCapabilities(), method);
         }
 
         boolean appPathProvided = StringUtils.isNotBlank(appPath);
@@ -202,7 +203,6 @@ public class MobileTestSession extends AbstractTestSession {
             this.platform = WebDriverPlatform.IOS;
         }
 
-        initializeAdditionalCapabilities(deviceTestAnnotation.additionalCapabilities(), method);
         logger.exiting();
     }
 
--- a/client/src/main/java/com/paypal/selion/platform/grid/browsercapabilities/ProxyHelper.java
+++ b/client/src/main/java/com/paypal/selion/platform/grid/browsercapabilities/ProxyHelper.java
@@ -12,7 +12,7 @@ import com.paypal.selion.configuration.Config.ConfigProperty;
  */
 final class ProxyHelper {
 
-    public static final String WARNING_MSG = "Enabling Proxy server settings on %s is known to change the proxy server settings at the machine Level!\n"
+    public static final String WARNING_MSG = "Enabling Proxy server settings on %s is known to change the proxy server settings at the machine Level!%n"
             + "So if your remote host supports concurrent test runs it can cause un-predictable test results!";
 
     private ProxyHelper() {
--- a/client/src/main/java/com/paypal/selion/platform/html/Table.java
+++ b/client/src/main/java/com/paypal/selion/platform/html/Table.java
@@ -111,13 +111,14 @@ public class Table extends AbstractElement {
                 return 1;
             }
 
-            if (allRows != null)
+            if (allRows != null) {
                 for (WebElement row : allRows) {
                     if (!row.findElements(By.xpath("td")).isEmpty()) {
                         break;
                     }
                     dataStartIndex++;
                 }
+            }
         }
 
         return dataStartIndex;
--- a/client/src/main/java/com/paypal/selion/reports/reporter/html/HtmlReporterListener.java
+++ b/client/src/main/java/com/paypal/selion/reports/reporter/html/HtmlReporterListener.java
@@ -300,10 +300,11 @@ public class HtmlReporterListener implements IReporter, IInvokedMethodListener {
                     // Attaching ralogId to each of the page title.
                     if ((logLine.getHref() != null) && (logLine.getHref().length() > 1)) {
                         htmllog = "<a href='../" + logLine.getHref() + "' title='" + logLine.getLocation() + "' >"
-                                + logLine.getMsg() + "</a>";
+                                + (StringUtils.isNotEmpty(htmllog) ? htmllog : "Page Source") + "</a>";
+                        
                     }
                     // Don't output blank message w/o any Href.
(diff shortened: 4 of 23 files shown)


client/src/main/java/com/paypal/selion/platform/grid/browsercapabilities/ProxyHelper.java
```diff
@@ -12,7 +12,7 @@
  */
 final class ProxyHelper {
 
-    public static final String WARNING_MSG = "Enabling Proxy server settings on %s is known to change the proxy server settings at the machine Level!\n"
+    public static final String WARNING_MSG = "Enabling Proxy server settings on %s is known to change the proxy server settings at the machine Level!%n"
```
mach6: Why is this changed from `\n` to `%n`?  Also will you please make this line end at 120 characters?

client/src/main/java/com/paypal/selion/platform/html/Table.java
```diff
@@ -111,7 +111,7 @@ public synchronized int getDataStartIndex() {
                 return 1;
             }
 
-            if (allRows != null)
+            if (allRows != null){
```
mach6: Please fix the formatting and indentation of this `if` block.

client/src/main/java/com/paypal/selion/testcomponents/AbstractPage.java
```diff
@@ -128,7 +128,7 @@ protected void loadObjectMap() {
                     objectMap = currentObjectMap;
                 }
             } catch (Exception e) {
-                throw new RuntimeException("Unable to initialize page data for " + pageDomain + "/" + pageClassName
+                throw new RuntimeException("Unable to initialize page data for " + pageDomain + "/" + pageClassName //NOSONAR
```
mach6: The `//NOSONAR` in the middle of the `new RuntimeException` looks odd. Will you please declare a new variable and pass that to the call?

client/src/main/java/com/paypal/selion/reports/reporter/services/ReporterDateFormatter.java
```diff
@@ -36,10 +36,12 @@
 
     private static SimpleLogger logger = SeLionLogger.getLogger();
 
-    // DEVNOTE: Format required to support Internet Explorer browser.
-    private static final DateFormat formatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX");
-
     public static final String CURRENTDATE = "currentDate";
+    
+    private static DateFormat getFormatter(){
```
mach6: Please fix the formatting on this method

client/src/main/java/com/paypal/selion/reports/reporter/runtimereport/JsonRuntimeReporterHelper.java
```diff
@@ -83,7 +83,8 @@
     public JsonRuntimeReporterHelper() {
         try {
             File workingDir = new File(Config.getConfigProperty(ConfigProperty.WORK_DIR));
-            workingDir.mkdirs();
+            boolean isCreated = workingDir.mkdirs();
+            logger.log(Level.FINER,"Is working directory creation successful:"+ isCreated);
```
mach6: Please reword this message and fix the formatting.

I suggest;

``` java
logger.log(Level.FINER, "Working directory created successfully: " + isCreated);
```

That said, we don't seem to be doing anything if it was not successfully created. Should we be? If not, then why bother logging the message?

client/src/main/java/com/paypal/selion/reports/runtime/SaverFileSystem.java
```diff
@@ -131,8 +131,8 @@ public PageContents getScreenshotByName(String name) throws IOException {
     @Override
     public void init() {
         logger.entering();
-        (new File(outputFolder)).mkdirs();
```
fakrudeen78, line 134: do we need to handle negative scenario? i.e.. what happens when the outputFolder directory is not created?

client/src/main/java/com/paypal/selion/platform/grid/browsercapabilities/ProxyHelper.java
ysecaran: @mach6 : Sonar highlighted this with the following reason:

This format string include a newline character (\n). In format strings, it is generally preferable better to use %n, which will produce the platform-specific line separator.

client/src/main/java/com/paypal/selion/reports/runtime/SaverFileSystem.java
ysecaran, line 134: let me validate this

client/src/main/java/com/paypal/selion/reports/reporter/runtimereport/JsonRuntimeReporterHelper.java
ysecaran: @mach6  yes, it is not used anywhere. I added this to overcome the violation and also it made sense to look this up in the log while debugging. I will reword this.

client/src/main/java/com/paypal/selion/reports/runtime/SaverFileSystem.java
ysecaran, line 134: @fakrudeen78 the outputfolder points to target/surefire-reports/ which gets created by surefire plugin. It is used as the base directory to store all the details of reports. This statement can be removed because the outputfolder is already created.

fakrudeen78, line 134: do we have same behaviour when we run locally through testng plugin?

ysecaran, line 134: In case of triggering the test case execution from eclipse the output folder will be {Base directory}\test-output\

I tested the same by deleting and commenting out the code to create directory in the init() method . The folder gets created and inside it the report resources and the actual report gets populated.

In both cases i ran the suite as well as a standalone test case. The output related to reports are getting populated properly

client/src/main/java/com/paypal/selion/reports/reporter/html/HtmlReporterListener.java
```diff
@@ -300,11 +300,13 @@ private String getContent(ITestResult result) {
                     // Attaching ralogId to each of the page title.
                     if ((logLine.getHref() != null) && (logLine.getHref().length() > 1)) {
                         htmllog = "<a href='../" + logLine.getHref() + "' title='" + logLine.getLocation() + "' >"
-                                + logLine.getMsg() + "</a>";
+                                 + "Page Source </a>";
```
mach6: This change introduces a breakage in the older Html Report. 

If I call;

``` java
SeLionReporter.log("foo",true,true)
```

The link text should be `foo`.. With the change, the link text will always be `Page Source`.

client/src/main/java/com/paypal/selion/platform/grid/MobileTestSession.java
```diff
@@ -131,7 +131,7 @@ public void initializeTestSession(InvokedMethodInformation method) {
         if (StringUtils.isNotBlank(getLocalConfigProperty(ConfigProperty.MOBILE_NODE_TYPE))) {
             mobileNode = getLocalConfigProperty(ConfigProperty.MOBILE_NODE_TYPE);
         }
-
+
```
mach6: please remove this whitespace

client/src/main/java/com/paypal/selion/platform/grid/browsercapabilities/ProxyHelper.java
mach6: ah yes.. thanks

client/src/main/java/com/paypal/selion/reports/reporter/html/HtmlReporterListener.java
ysecaran: The intention to hard code this is , irrespective of a message is present or null or empty it will be displayed below the page source link.

I have tested use cases of asserts failing with null , a message and an empty message.
For null the link text will be null, for empty there wont be a link at all. 
This does not break the report but the page source link will be presented irrespective of the message

client/src/main/java/com/paypal/selion/platform/grid/MobileTestSession.java
ysecaran: will fix this

client/src/main/java/com/paypal/selion/reports/reporter/html/HtmlReporterListener.java
mach6: ... and that is a change behavior and an inconsistency with how RuntimeReporter will render things. We need the original behavior to remain intact and to also address the sonar violation

ysecaran: @mach6 the fix is done. When the message is empty or null the link text will have page source

PR #135, opened 2015-07-29
Refactor DataProviders into standalone module

mach6:
This is PR #124 updated and pushed to a local branch 
- Refactor the SeLion DataProviders into a standalone module dataprovider
  that is independent of the rest of SeLion client and can be used by
  clients using TestNG.
- Moved SeLionLogger (as well as LoggerConfig) and FileAssistant into
  common as well as their associated tests.
- Moved dataprovider test resources into client-dataprovider.
- Moved to commons-configuration and commons-codec 1.10


server/pom.xml
```diff
@@ -98,8 +97,8 @@
         </dependency>
 
         <dependency>
-            <groupId>commons-lang</groupId>
-            <artifactId>commons-lang</artifactId>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
```
mach6: This change make little sense.. None of our code in `server` is using `lang3`

project-bom/pom.xml
```diff
@@ -100,6 +100,12 @@
                 <version>1.9.2</version>
             </dependency>
 
+            <dependency>
+                <artifactId>guava</artifactId>
+                <groupId>com.google.guava</groupId>
+                <version>18.0</version>
+            </dependency>
```
mach6: can be removed.. we get this dependency transitively

project-bom/pom.xml
```diff
@@ -321,6 +327,12 @@
                 <groupId>org.apache.velocity</groupId>
                 <artifactId>velocity</artifactId>
                 <version>1.7</version>
+                <exclusions>
```
mach6: this scares me.. `lang` and `lang3` have different namespaces.. I don't think `velocity` will let us swap one out for the other.

common/pom.xml
```diff
@@ -63,9 +62,15 @@
...
         <dependency>
-            <groupId>commons-lang</groupId>
-            <artifactId>commons-lang</artifactId>
+            <groupId>commons-configuration</groupId>
+            <artifactId>commons-configuration</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
```
mach6: Nothing in `common` uses this.

common/pom.xml
```diff
@@ -74,6 +79,21 @@
         </dependency>
 
         <dependency>
+            <groupId>commons-logging</groupId>
```
mach6: `commons-config` in our `common` module already pulls this in.

mach6:
@sebady I believe I have ironed out the kinks with the original PR .. Please review. I'm a little bit dissatisfied that we have `SeLionLogger` in `common` now as a part of this change -- since SeLion-Grid still has it's own `SeLionGridLogger`.. I don't think it's super pressing but perhaps we should try to merge the capabilities of each (`SeLionLogger` and `SeLionGridLogger`) into `common` in a future change set?
@fakrudeen78 @krmahadevan anything to add to this?

project-bom/pom.xml
```diff
@@ -31,7 +31,13 @@
             <dependency>
                 <groupId>commons-lang</groupId>
                 <artifactId>commons-lang</artifactId>
```
sebady, line 33: I think we can remove commons-lang entirely as a direct dependency and just use commons-lang3 going forward.

mach6, line 33: unfortunately that is not the case. we have dependencies which require both (one or the other) and they are not a interchangeable without code (that we don't own) updates  --- the `imports` are different.

PR #139, opened 2015-08-03
FIx #129 - Check port availability on local grid startup

### LINKED ISSUE #129: SeLion Grid does not get shutdown after run when using runLocally=true
After the change with running locally the server is not getting shutdown correctly when the run is finished. When I run another grid is getting started which at some points leads to multiple running grid's which use up all system resources.

screen shot 2015-07-23 at 8 46 04 am

mach6:


client/src/main/java/com/paypal/selion/platform/grid/AbstractBaseLocalServerComponent.java
```diff
@@ -91,10 +94,29 @@ public synchronized void shutdown() {
...
+                throw new IllegalArgumentException(portInUseError);
+            }
+        }
+    }
+
     /*
-     * Waits for the launcher to initialize.. This code exists purely to give the launcher more time to
-     * download dependencies. Otherwise, it is not needed since RunnableLauncher#isRunning() _should_ check for 
-     * isInitialized() at method entry.
+     * Waits for the launcher to initialize.. This code exists purely to give the launcher more time to download
```
sbheemanadham: There are two periods after the initialize

sbheemanadham:
I have completed verification. The pull request looks good except minor comment on the javadoc.

PR #142, opened 2015-08-10
Restructured Driver and DesiredCapability factory for Mobile & Web Te…

sundaramrajendran:
…st Session
- Created WebDriverFactory and MobileDriverFactory for Web and Mobile
- Removed DriverFactory and DesiredCapabilityFactory
- Removed deprecated IPad and IPhone browser flavors and corresponding classes


mach6:
@sundaramrajendran Can you get rid of `BrowserFlavors.GENERIC` too? Also, there are many changes here that increase the visibility of classes and/or methods to `public`. I'd like to avoid this. Visibility was at the `package` level before this change, by design.

sundaramrajendran:
@mach6 I have addressed review comments

mach6:
this is merged. thank you!

PR #147, opened 2015-08-12
Fix: improve the clean up of iexplore browser

sebady:
IExplore browsers were being missed from our process cleanup when
they have native dialog displayed.  The iexplore process would be
orphaned in this case and was not found during search of all our
children pids.  We now find and kill all iexplore browsers,
whether they are node process children or not.


sebady:
This is merged

PR #150, opened 2015-08-18
Load the Runtime reporter test case summary from JSON file

sundaramrajendran:


client/src/main/resources/templates/js/main.js
```diff
@@ -195,19 +195,20 @@ function displaySource(src, title) {
...
-      } else if (item.status === 'Skipped') {
-        skipped = skipped + 1;
-      } else {
-        running = running + 1;
-      }
-    });
+    if(reports.reportSummary.testMethodsSummary.passed) {
+      passed = reports.reportSummary.testMethodsSummary.passed;
+    }
+    if(reports.reportSummary.testMethodsSummary.failed) {
```
sebady: Would keeping the "else if" 's be preferable here.

sundaramrajendran: @sebady We can't, Because each if condition is not related each other

mach6:
@sundaramrajendran @sebady looks good to me

sebady:
merged

Issue #32, opened 2014-10-15
Add configuration setting to set selion log file limits.


mach6:
With the recent changes to `SimpleLogger` (which is used by `SeLionLogger`) https://github.com/paypal/SeLion/commit/003b48052acddf8e7c3ba7f7499cbe893eda8b46 we have an opportunity to add a new configuration option that will control the log file limits. For example, we can now limit the number of log files and/or the size before rolling to a new file. We should define a default behavior, add it to our `SeLionLoggerSettings`, and give the user the ability to override the default behavior.

ysecaran:
@mach6 : I am getting this done. Will send out a pull request after testing

mach6:
This was fixed with #151

PR #151, opened 2015-08-19
Fix #31 Allow users to specify log file size and count

ysecaran:
- Added new values to LoggerProperties enum
- utilized the new properties when initializing the logger


mach6:
@sebady @ysecaran this looks good to me. thank you!

mach6:
This is actually a fix for #32

PR #153, opened 2015-08-24
Fix #152: Code changes to reset the pagination when selecting treevie…

### LINKED ISSUE #152: RuntimeReporter TreeView selection does not reset pagination
To reproduce this issue;
1. View a result set which takes up multiple pages.
2. Goto the last page
3. Goto TreeView and select a result set which takes up fewer pages (1 for example)
4. Observe the report is refreshed and no test results are shown.
5. Goto the first page, and now the results are shown.

It seems the TreeView needs to reset to page 1 when something is selected.

sundaramrajendran:
…w in Runtime Reporter


sebady:
merged

PR #155, opened 2015-08-26
Failing test case in SeleniumGridListener.beforeInvocation() method w…

sundaramrajendran:
…hen pre validation or configuration got failed

### DIFF
--- a/client/src/main/java/com/paypal/selion/platform/grid/SeleniumGridListener.java
+++ b/client/src/main/java/com/paypal/selion/platform/grid/SeleniumGridListener.java
@@ -137,10 +137,13 @@ public class SeleniumGridListener implements IInvokedMethodListener, ISuiteListe
                     System.exit(1); // NOSONAR
                 }
             }
+        } catch (RuntimeException e) { // NOSONAR
+            Grid.getThreadLocalException().set(e);
         } catch (Exception e) { // NOSONAR
-            logger.log(Level.WARNING, "An error occurred while processing beforeInvocation: " + e.getMessage(), e);
+            //Must wrap the Exception to RuntimeException, So that users no need to
+            //add try catch block in their test case when invoke Grid.driver()
+            Grid.getThreadLocalException().set(new RuntimeException(e.getMessage(), e));
         }
-
         logger.exiting();
     }
 
--- a/client/src/main/java/com/paypal/selion/platform/grid/Grid.java
+++ b/client/src/main/java/com/paypal/selion/platform/grid/Grid.java
@@ -48,6 +48,7 @@ public final class Grid {
     private static SimpleLogger logger = SeLionLogger.getLogger();
     private static ThreadLocal<RemoteWebDriver> threadLocalWebDriver = new ThreadLocal<RemoteWebDriver>();
     private static ThreadLocal<AbstractTestSession> threadTestSession = new ThreadLocal<AbstractTestSession>();
+    private static ThreadLocal<RuntimeException> threadLocalException = new ThreadLocal<RuntimeException>();
 
     static {
         Logger.getLogger("").setLevel(Level.OFF);
@@ -63,6 +64,10 @@ public final class Grid {
 
     static ThreadLocal<AbstractTestSession> getThreadLocalTestSession() {
         return threadTestSession;
+    }
+
+    static ThreadLocal<RuntimeException> getThreadLocalException() {
+        return threadLocalException;
     }
 
     /**
@@ -84,6 +89,12 @@ public final class Grid {
      *         {@link WebTest} flow.
      */
     public static RemoteWebDriver driver() {
+        //Need to throw an RuntimeException if already caught on SeleniumGridListener.beforeInvocation()
+        RuntimeException runtimeException = threadLocalException.get();
+        if(runtimeException != null) {
+            throw runtimeException;
+        }
+
         AbstractTestSession testSession = getTestSession();
         if (!testSession.isStarted()) {
             testSession.startSesion();


client/src/main/java/com/paypal/selion/platform/grid/SeleniumGridListener.java
```diff
@@ -125,6 +126,8 @@ public void beforeInvocation(IInvokedMethod method, ITestResult testResult) {
             }
         } catch (Exception e) { // NOSONAR
             logger.log(Level.WARNING, "An error occurred while processing beforeInvocation: " + e.getMessage(), e);
+            //Failing the test case, because pre validation/configuration got failed
+            Assert.fail(e.getMessage(), e);
```
krmahadevan: @sundaramrajendran 

IMO, `Asserts` are to be used ONLY in tests and Not within a framework.

mach6: @sundaramrajendran I'm in agreement with @krmahadevan.
The primary issue with this approach is the failure message will be associated with the class `SeleniumGridListener`

Is there a more elegant way of doing this? What happens if you re-throw the `Exception` to TestNG?

sundaramrajendran: @krmahadevan @mach6 : I agree, let me verify any other option available to address this...

sundaramrajendran:
Please review this

client/src/main/java/com/paypal/selion/platform/grid/SeleniumGridListener.java
```diff
@@ -123,10 +123,11 @@ public void beforeInvocation(IInvokedMethod method, ITestResult testResult) {
                     System.exit(1);  // NOSONAR
                 }
             }
+        } catch (RuntimeException e) { // NOSONAR
+            Grid.getThreadLocalException().set(e);
         } catch (Exception e) { // NOSONAR
```
mach6, line 142: This PR does not fix the issue when a checked exception occurs in `beforeInvocation`.. Since a message is only logged when an `Exception` is caught,`@WebTest` and/or `@MobileTest` execution continues and a `NullPointerException` will happen in the first call into `Grid`.. Furthermore, the true failure (the original `Exception`), is not included with the test results.

mach6:
@sundaramrajendran I like this approach better than before but as I mention it does not fully address the issue.

sundaramrajendran:
@mach6 I have added same fix to Exception block also. But wrapping to RuntimeException, because users don't need to add try catch block in their test case when invoke Grid.driver()

client/src/main/java/com/paypal/selion/platform/grid/SeleniumGridListener.java
```diff
@@ -137,10 +137,13 @@ public void beforeInvocation(IInvokedMethod method, ITestResult testResult) {
                     System.exit(1); // NOSONAR
                 }
             }
+        } catch (RuntimeException e) { // NOSONAR
+            Grid.getThreadLocalException().set(e);
         } catch (Exception e) { // NOSONAR
-            logger.log(Level.WARNING, "An error occurred while processing beforeInvocation: " + e.getMessage(), e);
+            //Must wrap the Exception to RuntimeException, So that users no need to
+            //add try catch block in their test case when invoke Grid.driver()
+            Grid.getThreadLocalException().set(new RuntimeException(e.getMessage(), e));
```
sebady, line 145: Why not convert to RuntimeException at the throw site instead?

Also what if multiple exceptions have occurred in BeforeX config methods before Grid.driver().  We are saving only the last and clobbering all earlier ones.

sundaramrajendran, line 145: @sebady 
a) Why not convert to RuntimeException at the throw site instead?
     Purpose is to catch all the exception, should not leak any exception. Already code validation and all the custom exceptions are RuntimeException only.

b) Also what if multiple exceptions have occurred in BeforeX config methods before Grid.driver(). We are saving only the last and clobbering all earlier ones.?
       Scenario1: Our logic in SeleniumGridListener.beforeInvocation() wont execute at all for the BeforeX config methods in Non session sharing test. So that no exception will be occurred in belforeInvocation method. If some code problem in BeforeX methods then subsequent test method will be skipped.
       Scenario2: Logic in SeleniumGridListener.beforeInvocation() method will be executed for BeforeClass method in Session sharing test. If something went wrong then exception will be saved and will be thrown when Grid.driver() invoked in BeforeClass method itself. So subsequent test case will be skipped. 

I have tried to reproduce the scenario you are talking about here. But I am not able to reproduce this issue by having valid scenario. In case more than one issues are occurred then latest exception only thrown to the user. Once user solve the latest issue then user will get previous exception for the second run. Is it fine?

sebady, line 145: What I was recommending for your recent addition was to also start using this instead:
private static ThreadLocal<Exception> threadLocalException = new ThreadLocal<Exception>();

I thought that in the sharing case we could have several exceptions in beforeX but with this recent addition in beforeInvocation:

if (Grid.getTestSession() != null) {
                return;
 }

That is not possible now.

So ignore my 2nd comment:

mach6:
@sebady are all of your concerns addressed?

sebady:
I think it's a little cleaner to save the exception AS-IS in thread local and convert at the throw site to runtimeException or whatever.  This is pretty minor though.

mach6:
@sebady making your suggested change to this and merging it.
@sundaramrajendran fyi

mach6:
Updated and merged this.

PR #157, opened 2015-09-01
Fix: Fix process case name on mac/unix.

sebady:
We were not matching on the right case on mac/unix in ProcessHandler.
Also a minor change to throw illegal state exception for Grid.Driver()
in a method missing a webstest.

### DIFF
--- a/client/src/main/java/com/paypal/selion/platform/grid/Grid.java
+++ b/client/src/main/java/com/paypal/selion/platform/grid/Grid.java
@@ -81,10 +81,15 @@ public final class Grid {
      * @return A {@link RemoteWebDriver} object which can be used in the case of {@link MobileTest} and {@link WebTest}s
      */
     public static RemoteWebDriver driver() {
-        if (! getTestSession().isStarted()) {
-            getTestSession().startSesion();
+        AbstractTestSession testSession = getTestSession();
+        if (! testSession.isStarted()) {
+            testSession.startSesion();
         }
-        return threadLocalWebDriver.get();
+        RemoteWebDriver rwd = threadLocalWebDriver.get();
+        if (rwd == null) {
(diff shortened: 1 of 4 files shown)


client/src/main/java/com/paypal/selion/platform/grid/Grid.java
```diff
@@ -81,10 +81,15 @@ public static long getExecutionTimeoutValue() {
...
     public static RemoteWebDriver driver() {
-        if (! getTestSession().isStarted()) {
-            getTestSession().startSesion();
+        AbstractTestSession testSession = getTestSession();
+        if (! testSession.isStarted()) {
+            testSession.startSesion();
+        }
+        RemoteWebDriver rwd = threadLocalWebDriver.get();
+        if (rwd == null) {
+            throw new IllegalStateException("Driver not initialized. Is @WebTest annotation missing on method?");
```
mach6: or `@MobileTest`.. If the user is running session sharing, the annotation will be on the class vs. the method.

server/src/main/java/com/paypal/selion/utils/process/AbstractProcessHandler.java
```diff
@@ -72,7 +72,7 @@
                     tProcess = new ProcessInfo(eachProcessData[1], eachProcessData[2]);
                     break;
                 }
-                if (matches(tProcess.getProcessName())) {
+                if (matches(tProcess.getProcessName().toLowerCase())) {
```
mach6: Since we are already formatting the `processName` in https://github.com/paypal/SeLion/blob/develop/server/src/main/java/com/paypal/selion/pojos/ProcessInfo.java, can we just do this there?

mach6:
:+1: LGTM. Will merge in a bit

PR #160, opened 2015-09-02
Added LICENSE to every SeLion jar file. Closes #108

### LINKED ISSUE #108: Add LICENSE file to resulting jars
The jar files should have a copy of the LICENSE file in them.

sundaramrajendran:


mach6:
@sundaramrajendran What about the archetype and codegen artifacts? How about using maven to copy/create these resources instead of having multiple copies?

sundaramrajendran:
@mach6 Addressed review comments

mach6:
@sundaramrajendran this looks good to me.. The only change I'll suggest is to use `src/main/resources` everywhere.. You have mixed usage of `src\main\resources` and `src/main/resources`. Note: use of backslash vs. forwardslash

sundaramrajendran:
@mach6 Done

PR #163, opened 2015-09-02
Fix #156 - Allow access to Grid session in `@BeforeClass` and `@After…

### LINKED ISSUE #156: Grid is Null when use Grid in @Beforexxx
As I see the SeleniumGridListener, there is a method to check the invoked method is test method or not, but I think there are something missing. When we use the currently session or create a new session in the @BeforeXXX methods, it won't create a new session.

This situation will be used for some common actions before the test case method.

Hope it will be fixed later.

### LINKED ISSUE #158: method parameter shown as 'undefined' in RuntimeReporter html output
When you view the 'Reporter Output' or the 'Stacktrace', in the RuntimeReporter html output, if the test method has no parameters, the UI shows the method parameters as `com.foo.bar.TestClass.testMethod(undefined)`

Instead, it should be `com.foo.bar.TestClass.testMethod()`

mach6:
…Class` when session sharing.

Also in this change set;
- Fix #158 - guard against `undefined` in `RuntimeReporter` html report
- Added support for 'Reporter Output' in configuration methods on `RuntimeReporter` html report
- Add more debug messages to `DebugListner`

### DIFF
--- a/client/src/main/java/com/paypal/selion/platform/grid/SeleniumGridListener.java
+++ b/client/src/main/java/com/paypal/selion/platform/grid/SeleniumGridListener.java
@@ -90,22 +90,36 @@ public class SeleniumGridListener implements IInvokedMethodListener, ISuiteListe
                 return;
             }
 
-            if (!method.isTestMethod()) {
+            // For non-session sharing, we only allow our annotation(s) on @Test methods.
+            // When this condition is true, we allow the session be created.
+            if (!method.isTestMethod() && !isSeLionAnnotatedTestClass(method)) {
                 return;
             }
 
-            Class<?> cls = method.getTestMethod().getInstance().getClass();
-            boolean isWebTestClass = cls.getAnnotation(WebTest.class) != null;
-            boolean isMobileTestClass = cls.getAnnotation(MobileTest.class) != null;
-
-            if ((isWebTestClass || isMobileTestClass)) {
-                if (isLowPriority(method)) {
-                    // For session sharing tests, Need to create new session only for Test (Web or Mobile) with lowest
-                    // priority (first test) in the class.
-                    testSessionSharingRules(method);
-                } else {
+            // For session sharing, we only allow our annotation on the class.
+            // In this case the session can only be created in the @Test with the highest priority (first test, smallest
+            // number) or in a @BeforeClass
+            if (isSeLionAnnotatedTestClass(method)) {
+                if (!isValidBeforeCondition(method)) {
                     return;
                 }
+
+                // For session sharing, we need to ensure @Test methods are priority based.
+                if (method.isTestMethod()) {
+                    if (isLowPriority(method)) {
+                        // For session sharing tests, Need to create new session only for Test (Web or Mobile) with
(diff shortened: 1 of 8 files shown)


mach6:
@sebady @sundaramrajendran Please review this.

client/src/main/resources/templates/js/main.js
```diff
@@ -601,7 +601,7 @@ function displaySource(src, title) {
     $('.btn-screenshot' + buttonAdditonalClassName).click(function() {
       var testcase = data[parseInt($(this).attr('data-index'))];
       var params = ' ';
-      if (testcase.parameters !== null) {
+      if ((testcase.parameters !== null) && (typeof testcase.parameters !== 'undefined')) {
```
sundaramrajendran: We can simplify the condition as if (testcase.parameters) will validate the null check and undefined check

sundaramrajendran:
@mach6 Everything looks good except one comment as posted

client/src/test/java/com/paypal/selion/platform/grid/SessionSharingTestWithConfigurationMethods.java
```diff
@@ -0,0 +1,124 @@
...
+@WebTest(browser = "chrome")
+@Test(singleThreaded = true)
+public class SessionSharingTestWithConfigurationMethods {
+    @BeforeSuite
+    public void beforeSuite() {
+        // not started yet
+        Assert.assertNull(Grid.getTestSession());
+    }
+
+    @BeforeTest
```
ILikeToNguyen, line 48: Did you try multiple @BeforeTest, multiple etc?  The logic looks right in SeleniumGridListener but just wanted to double check.

client/src/main/java/com/paypal/selion/platform/grid/SeleniumGridListener.java
```diff
@@ -90,22 +90,41 @@ public void beforeInvocation(IInvokedMethod method, ITestResult testResult) {
...
-            if ((isWebTestClass || isMobileTestClass)) {
-                if (isLowPriority(method)) {
-                    // For session sharing tests, Need to create new session only for Test (Web or Mobile) with lowest
-                    // priority (first test) in the class.
-                    testSessionSharingRules(method);
-                } else {
+            // For session sharing, we only allow our annotation on the class.
+            // In this case the session can only be created in the @Test with the highest priority (first test, smallest
+            // number) or in a @BeforeClass
+            if (isSeLionAnnotatedTestClass) {
```
sebady: The PR look good.

There is some opportunity for refactoring here.  Since all we need to have here is "isSeLionAnnotatedTestClass" we can put that computation in a private method

client/src/test/java/com/paypal/selion/platform/grid/SessionSharingTestWithConfigurationMethods.java
ILikeToNguyen, line 48: Disregard this.  I thought the session would be available during before and after annotations but they are not by design.

PR #166, opened 2015-09-07
Infer Criteria from the current ManagedArtifact.

mach6:
- Removed userId from interface Criteria.
- Moved DefaultCriteria into its own file.
- DefaultManagedArtifact implements ManagedArtifact<DefaultCriteria>.
- Requested Criteria is processed from HTTP path by DefaultManagedArtifact.
  This was earlier processed by DownloadResponder.
- Removed 'managedCriteria' configuration from SeLionConfig.json.
- Revised interface ManagedArtifact.
- Interface ServerRepository updated to process HTTP path information instead of Criteria.
- Updated unit tests and documentations.


server/src/main/java/com/paypal/selion/grid/servlets/transfer/Criteria.java
```diff
@@ -34,14 +34,6 @@
     String getArtifactName();
 
     /**
-     * Returns the user Id this artifact is tied to. This is essentially the folder by the name of the user that houses
-     * the artifact.
-     * 
-     * @return User Id of the criteria.
-     */
-    String getUserId();
```
mach6: The javadoc for the class is out of date due to this change.

server/src/main/java/com/paypal/selion/grid/servlets/transfer/UploadRequestProcessor.java
```diff
@@ -53,7 +53,7 @@
     /**
      * @return a {@link List} of {@link ManagedArtifact} which represent items on the {@link ManagedArtifactRepository}
```
mach6: @shelar javadoc needs update as a result of this change

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultCriteria.java
```diff
@@ -0,0 +1,119 @@
...
+    private String artifactName;
+
+    private String userId;
+
+    private String applicationFolder;
+
+    public DefaultCriteria(EnumMap<RequestHeaders, String> parametersMap) {
+        validateParametersMap(parametersMap);
+        this.artifactName = parametersMap.get(RequestHeaders.FILENAME);
+        this.userId = parametersMap.get(RequestHeaders.USERID);
```
mach6: @shelar We seem have made the criteria for the artifact more flexible from a storage and download perspective but still have a limitation on the upload front. Case in point, see enum `UploadRequestProcessor#RequestHeaders`. Do you agree?

server/src/main/java/com/paypal/selion/grid/servlets/transfer/ServerRepository.java
```diff
@@ -26,7 +26,7 @@
  * @param <U>
```
mach6: @shelar this javadoc is out of date due to this change set

server/src/main/java/com/paypal/selion/grid/servlets/transfer/Criteria.java
mach6: @shelar After further thought, I think we should revert this change. A `userId` is a basic condition that a `ManagedArtifact` must conform to. It allows for the server repository to be multi-tenant.

server/src/main/java/com/paypal/selion/grid/servlets/transfer/ManagedArtifact.java
```diff
@@ -57,7 +57,7 @@
      *            Instance of {@link Criteria} to match.
      * @return true if there is a match, false otherwise.
      */
-    <U extends Criteria> boolean matches(U criteria);
+    boolean matchesCriteria(String pathInfo);
```
mach6: @shelar I think this should be called `matchesPathInfo`

server/src/main/java/com/paypal/selion/grid/servlets/transfer/ManagedArtifact.java
```diff
@@ -20,7 +20,7 @@
  * call.
  * 
  */
-public interface ManagedArtifact {
+public interface ManagedArtifact<U extends Criteria> {
```
mach6: @shelar With the other changes you have made, I think we should be able to avoid turning this into a template type

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultCriteria.java
squash-merge: I agree, all upload side is limited by headers. I see how it defeats the purpose of the changes I submitted. Maybe, I should have modified `UploadedArtifact` to contain all header name-values in a property, and let `ServerRepository#saveContents` use it.

mach6:
@shelar @sebady @sundaramrajendran @kumaravel-jayakumar Please take a look at this PR. Thanks!

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultCriteria.java
```diff
@@ -0,0 +1,118 @@
...
+    private String artifactName;
+
+    private String uuidFolderName;
+
+    private String subFolderName;
+
+    public DefaultCriteria(Map<String, String> parametersMap) {
+        validateParametersMap(parametersMap);
+        this.artifactName = parametersMap.get(ManagedArtifact.ARTIFACT_FILE_NAME);
+        this.uuidFolderName = parametersMap.get(DefaultRequestParameters.USER_ID);
```
sebady: is uuid field a uuid or a "user id"?

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultManagedArtifact.java
```diff
@@ -48,29 +47,95 @@
...
+        public boolean isRequired(String parameter) {
+            if (getParameters().containsKey(parameter)) {
+                return getParameters().get(parameter).booleanValue();
+            }
+            return false;
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    public <T extends RequestParameters> T getRequstParameters() {
```
squash-merge: Should `getRequstParameters` be changed to `getRequestParameters`?

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultCriteria.java
```diff
@@ -0,0 +1,118 @@
...
+    public String getArtifactFolder() {
+        return subFolderName;
+    }
+
+    public Map<String, String> asMap() {
+        LOGGER.entering();
+        Map<String, String> contentMap = new HashMap<>();
+        contentMap.put(ManagedArtifact.ARTIFACT_FILE_NAME, getArtifactName());
+        contentMap.put(DefaultRequestParameters.USER_ID, getUUIDFolderName());
+        if (!StringUtils.isBlank(getArtifactFolder())) {
```
sebady: This if test is not really necessary

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultCriteria.java
```diff
@@ -0,0 +1,118 @@
...
+        }
+        if (!getUUIDFolderName().equals(otherCriteria.getUUIDFolderName())) {
+            return false;
+        }
+        boolean equals = getArtifactFolder() == null ? otherCriteria.getArtifactFolder() == null
+                : getArtifactFolder().equals(otherCriteria.getArtifactFolder());
+        if (equals == false) {
+            return false;
+        }
+        return true;
```
sebady: Why not just return equals?

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultManagedArtifact.java
```diff
@@ -48,29 +47,95 @@
...
 
-    private String parentFolderName = null;
+    private String uuidFolderName = null;
 
     private byte[] contents = null;
 
-    private final long timeToLiveInMillis;
+    private static final long timeToLiveInMillis = ConfigParser.parse().getLong(EXPIRY_CONFIG_PROPERTY);
+
+    private String requestPathInfo;
```
sebady: Please add a comment that makes it clearer what this field  represents.  Is this the same as the full path?

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultCriteria.java
mach6: Well, it's really a "uid".. Nothing about the intent requires the identifier to be associated with a user. It just needs to be a unique identifier. Probably the request parameter should be changed to reflect this too. I left that part in tact, as you can see.

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultManagedArtifact.java
mach6: good catch. will fix

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultCriteria.java
mach6: This was pre-existing and I believe the intent was to check for equality on two instances of `DefaultCriteria` which are pointing to the same meta info. Honestly, since `Criteria` only come into play for download means, I don't think this is needed.

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultCriteria.java
mach6: This was pre-existing and I believe the if check is necessary, since `artifactFolder` is an optional parameter in the `DefaultManagedArtifact` implementation, though I need to see how `asMap()` is used.

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultManagedArtifact.java
mach6: Will do. It's the requested path to the artifact at download time. It is not the file system absolute path.

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultManagedArtifact.java
```diff
@@ -48,29 +47,95 @@
...
+    }
+
+    public void initFromPath(String pathName) {
+        // do not allow path that are empty or try to break out of the repository folder
+        Preconditions.checkArgument(StringUtils.isNotBlank(pathName));
+        Preconditions.checkArgument(!StringUtils.contains(pathName, ".."));
+
+        String repoPath = ManagedArtifactRepository.getInstance().getRepositoryFolder().getAbsolutePath();
+        String filePath = pathName;
+        if (!pathName.contains(repoPath)) {
```
mach6: This check can fail on windows where the `repoPath` is on a different drive than the drive included in `pathName`.. Will fix this.

server/src/main/java/com/paypal/selion/grid/servlets/transfer/ManagedArtifact.java
```diff
@@ -30,37 +70,39 @@
...
 
     /**
-     * Returns the name of the folder this artifact is housed.
+     * Returns the name of the folder this artifact is housed. This folder is relative to the {@link ServerRepository}
+     * root folder
      * 
      * @return Folder name of this artifact
      */
-    String getFolderName();
+    String getArtifactFolderName();
```
mach6: can be removed from the interface. can be deduced from `getPath()`. no more callers outside of `@Test`s also

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultManagedArtifact.java
```diff
@@ -48,29 +49,90 @@
...
 
-    public DefaultManagedArtifact(String pathName) {
-        this.filePath = pathName;
-        artifactFile = new File(this.filePath);
-        timeToLiveInMillis = ConfigParser.parse().getLong(EXPIRY_CONFIG_PROPERTY);
+    static final class DefaultRequestParameters implements RequestParameters {
+        public static final String UID = "uid";
+
+        public Map<String, Boolean> getParameters() {
+            Map<String, Boolean> headers = new HashMap<>();
```
squash-merge: We can avoid creating a new HashMap on every call here.

server/src/main/java/com/paypal/selion/grid/servlets/transfer/UploadRequestProcessor.java
```diff
@@ -295,40 +289,38 @@ private int parseRequestAsFileItems() throws FileUploadException {
         }
 
         private void populateHeadersMap() {
-            EnumMap<RequestHeaders, String> headersMap = getMapFromHeaders();
+            Map<String, String> headersMap = getRequestHeadersMap();
```
squash-merge, line 297: SInce an ManagedArtifact's `instance` is created on the next line, could we change this code to use a single `instance`? 

```
ManagedArtifact instance = getManagedArtifactInstance();
Map<String, String> headersMap = getRequestHeadersMap(instance);
```

Corresponding modification to `getRequestHeadersMap()` would then be required.

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultManagedArtifact.java
```diff
@@ -185,109 +250,11 @@ private void readContents() {
...
-            result = 31 * result + (this.getApplicationFolder() != null ? this.getApplicationFolder().hashCode() : 0);
-            return result;
-        }
-
-        @Override
-        public String toString() {
-            return "[ artifactName: " + getArtifactName() + ", userId: " + getUserId() + ", applicationFolder: "
-                    + getApplicationFolder() != null ? getApplicationFolder() : "" + " ]";
+    public String getAbsolutePath() {
+        if (artifactFile == null) {
```
squash-merge: This 4 lines be made into one line.

server/src/main/java/com/paypal/selion/grid/servlets/transfer/DefaultManagedArtifact.java
```diff
@@ -100,38 +177,26 @@ public String getParentFolderName() {
...
-            LOGGER.exiting(false);
-            return false;
-        }
-        if (isApplicationFolderRequested(criteria) && applicationFolderAndUserIdMatches(criteria)) {
-            LOGGER.exiting(true);
-            return true;
-        }
-        boolean matches = !isApplicationFolderRequested(criteria) && userIdMatches(criteria);
+    public boolean matchesPathInfo(String pathInfo) {
+        LOGGER.entering();
```
squash-merge: `LOGGER.entering(pathInfo)` could be useful.

server/src/main/java/com/paypal/selion/grid/servlets/transfer/ManagedArtifactRepository.java
```diff
@@ -120,43 +113,15 @@ public ManagedArtifact saveContents(UploadedArtifact uploadedArtifact) {
...
+    public ManagedArtifact getArtifact(String pathInfo) {
         /*
          * This method does not guarantee the presence of an artifact after isArtifactPresent() is called on an
          * artifact. This is because the cleaner thread could have deleted a expired artifact in between the calls.
          */
         try {
             repositorySynchronizationLock.lock();
-            LOGGER.entering(requestedCriteria);
-            ManagedArtifact managedArtifact = getMatch(requestedCriteria);
+            LOGGER.entering();
```
squash-merge: `LOGGER.entering(pathInfo)` would be useful.

squash-merge:
Thats all from me.  @sebady @sundaramrajendran  @kumaravel-jayakumar Your comments please!

server/src/main/java/com/paypal/selion/grid/servlets/transfer/UploadRequestProcessor.java
mach6, line 297: I took a different approach to address this and assigned the result of calling `getManagedArtifactInstance().getRequestParameters()` to a member variable which is used throughout `UploadRequestProcessor`

PR #168, opened 2015-09-10
Fix #167 and other misc changes

### LINKED ISSUE #167: ClassCastException occurs when connected node does not use SeLionRemoteProxy
When there is a node which does not use `SeLionRemoteProxy` connected to a SeLion enhanced hub/server a `ClassCastException` will occur in `GridForceRestartDelegateServlet` and `GridAutoUpgradeDelegateServlet`

mach6:
Additional changes include;
- Correct some javadoc issues
- Remove checks for Grid.driver() != null.. This is not a valid
  condition anymore
- Remove dead code in SeLionReporter
- Add missing ListenerManager check to RuntimeReporterListner.onStart

### DIFF
--- a/client/src/main/java/com/paypal/selion/platform/utilities/WebDriverWaitUtils.java
+++ b/client/src/main/java/com/paypal/selion/platform/utilities/WebDriverWaitUtils.java
@@ -26,7 +26,6 @@ import org.openqa.selenium.support.ui.ExpectedConditions;
 import org.openqa.selenium.support.ui.WebDriverWait;
 
 import com.google.common.base.Preconditions;
-import com.paypal.selion.annotations.WebTest;
 import com.paypal.selion.logger.SeLionLogger;
 import com.paypal.selion.platform.grid.Grid;
 import com.paypal.selion.platform.html.support.HtmlElementUtils;
@@ -39,8 +38,6 @@ import com.paypal.test.utilities.logging.SimpleLogger;
 public final class WebDriverWaitUtils {
     
     private final static SimpleLogger logger = SeLionLogger.getLogger();
-    private static final String INVALID_STATE_ERR_MSG = "Please use this API only from within a @"
-            + WebTest.class.getSimpleName() + " annotated test method.";
 
     private WebDriverWaitUtils(){
         //Utility class. So hide the constructor to defeat instantiation
@@ -67,7 +64,6 @@ public final class WebDriverWaitUtils {
      */
     public static void waitUntilElementIsClickable(final String elementLocator) {
         logger.entering(elementLocator);
-        Preconditions.checkState(Grid.driver() != null, INVALID_STATE_ERR_MSG);
         By by = HtmlElementUtils.resolveByType(elementLocator);
         ExpectedCondition<WebElement> condition = ExpectedConditions.elementToBeClickable(by);
         waitForCondition(condition);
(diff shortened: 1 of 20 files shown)


mach6:
@sebady @sundaramrajendran Please review

client/src/main/java/com/paypal/selion/platform/utilities/WebDriverWaitUtils.java
```diff
@@ -67,7 +64,6 @@ private static void waitForCondition(ExpectedCondition<?> condition, long timeou
      */
     public static void waitUntilElementIsClickable(final String elementLocator) {
         logger.entering(elementLocator);
-        Preconditions.checkState(Grid.driver() != null, INVALID_STATE_ERR_MSG);
```
sebady, line 70: Why not still do a Grid.driver() here just to verify that the driver is initialized to the state we expect here.  Is there another way to confirm the driver is initialized here and in other places?

mach6, line 70: The contract for `Grid.driver()` states it will not return `null`. Thus the check is not needed. If you want to check that the `TestSession` is valid, that's a different meaning and is accessible via `Grid.getTestSession()`

server/src/main/java/com/paypal/selion/grid/servlets/GridAutoUpgradeDelegateServlet.java
```diff
@@ -114,21 +117,34 @@ protected void process(HttpServletRequest request, HttpServletResponse response)
...
+                    // TODO :: Address the assumption here that any node which uses SeLionRemoteProxy also has
+                    // NodeAutoUpgradeServlet available. Without truth in this assumption, the call to
+                    // proxy.release() will do nothing.
+                    try {
+                        SeLionRemoteProxy proxy = (SeLionRemoteProxy) eachProxy;
+                        if (!proxy.release(downloadJSON)) {
+                            pendingProxy.add(proxy.getId());
+                        }
+                    } catch (ClassCastException e) {
+                        // ignore, node is not likely compatible with NodeAutoUpgradeServlet
```
sundaramrajendran: @mach6 I think we can log the exception, because it will helpful when debugging.

server/src/main/java/com/paypal/selion/grid/servlets/GridAutoUpgradeDelegateServlet.java
```diff
@@ -114,21 +117,34 @@ protected void process(HttpServletRequest request, HttpServletResponse response)
...
+                    // proxy.release() will do nothing.
+                    try {
+                        if (!eachId.trim().isEmpty()) {
+                            SeLionRemoteProxy proxy = (SeLionRemoteProxy) getRegistry().getProxyById(eachId.trim());
+                            if ((proxy != null) && (!proxy.release(downloadJSON))) {
+                                pendingProxy.add(proxy.getId());
+                            }
                         }
+                    } catch (ClassCastException e) {
+                        // ignore, node is not likely compatible with NodeAutoUpgradeServlet
```
sundaramrajendran: @mach6 Here too, log the exception applicable places

server/src/main/java/com/paypal/selion/grid/servlets/GridAutoUpgradeDelegateServlet.java
```diff
@@ -114,21 +117,34 @@ protected void process(HttpServletRequest request, HttpServletResponse response)
...
-                        pendingProxy.add(proxy.getId());
+                    // TODO :: Address the assumption here that any node which uses SeLionRemoteProxy also has
+                    // NodeAutoUpgradeServlet available. Without truth in this assumption, the call to
+                    // proxy.release() will do nothing.
+                    try {
+                        SeLionRemoteProxy proxy = (SeLionRemoteProxy) eachProxy;
+                        if (!proxy.release(downloadJSON)) {
+                            pendingProxy.add(proxy.getId());
+                        }
+                    } catch (ClassCastException e) {
```
sundaramrajendran: @mach6 Instead of catching ClassCastException, We can add if condition to verify is this instance of SeLionRemoteProxy and execute the code. Else condition log the information for user about "Node incompatible for AutoUpgrade"

mach6:
@sundaramrajendran thanks for the feedback. I made the changes you suggested

PR #175, opened 2015-09-22
Fix generics due to TestNG and Appium upgrades

mach6:


mach6:
@sebady let me know if you have any concerns with this one.

client/src/main/java/com/paypal/selion/platform/grid/browsercapabilities/MobileDriverFactory.java
```diff
@@ -98,7 +98,7 @@ private static AndroidDriver createAppiumAndroidInstance(URL url, DesiredCapabil
         }
     }
 
-    private static IOSDriver createAppiumIOSInstance(URL url, DesiredCapabilities capability) {
+    private static IOSDriver<?> createAppiumIOSInstance(URL url, DesiredCapabilities capability) {
```
sebady: The wildcard return type here and for android are the only ones I'm not sure about here.  Technically it's a better practice to more narrowly bound this type if we can. For example AndroidDriver<? extends WebElement>

mach6: Good catch. Will take a look.

mach6:
this is merged

PR #182, opened 2015-09-30
Add timeout capability to NodeRecycleThread

mach6:
When specified, times out the NodeRecycleThread and
shuts down the node at the given value (in seconds).

### DIFF
--- a/server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
+++ b/server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
@@ -15,17 +15,14 @@
 
 package com.paypal.selion.proxy;
 
-import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
-import java.text.MessageFormat;
 import java.util.ArrayList;
-import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Level;
 
-import org.apache.commons.io.FileUtils;
+import org.apache.http.HttpResponse;
+import org.apache.http.HttpStatus;
 import org.apache.http.NameValuePair;
 import org.apache.http.client.ClientProtocolException;
 import org.apache.http.client.config.RequestConfig;
@@ -39,13 +36,17 @@ import org.openqa.grid.internal.Registry;
 import org.openqa.grid.internal.TestSession;
 import org.openqa.grid.selenium.proxy.DefaultRemoteProxy;
 
+import com.paypal.selion.SeLionBuildInfo;
 import com.paypal.selion.pojos.SeLionGridConstants;
+import com.paypal.selion.SeLionBuildInfo.SeLionBuildProperty;
+import com.paypal.selion.node.servlets.NodeForceRestartServlet;
 import com.paypal.selion.grid.servlets.GridAutoUpgradeDelegateServlet;
 import com.paypal.selion.logging.SeLionGridLogger;
 import com.paypal.selion.node.servlets.NodeAutoUpgradeServlet;
-import com.paypal.selion.node.servlets.NodeForceRestartServlet;
 import com.paypal.selion.utils.ConfigParser;
 import com.paypal.selion.utils.ConfigParser.ConfigParserException;
+import com.paypal.test.utilities.logging.SimpleLogger;
+import com.paypal.test.utilities.logging.SimpleLoggerSettings;
 
 /**
  * This is a customized {@link DefaultRemoteProxy} for SeLion. This proxy when injected into the Grid, starts counting
@@ -62,13 +63,13 @@ import com.paypal.selion.utils.ConfigParser.ConfigParserException;
 public class SeLionRemoteProxy extends DefaultRemoteProxy {
 
     private static final SeLionGridLogger LOGGER = SeLionGridLogger.getLogger(SeLionRemoteProxy.class);
+    private static SimpleLogger proxyLogger;
     private static final int MAX_SESSION_ALLOWED = 50;
 
     private int maxSessionsAllowed, totalSessionsCompleted = 0, totalSessionsStarted = 0;
     private boolean forceShutDown = false;
     private String machine;
-    private File logFile = null;
-    private NodeRecycleThread nodeRecycleThread = new NodeRecycleThread();
+    private NodeRecycleThread nodeRecycleThread = new NodeRecycleThread(getId());
 
     private int getUniqueSessionCount() {
         try {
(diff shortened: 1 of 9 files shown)


mach6:
@sebady @sundaramrajendran thoughts on this?

server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
```diff
@@ -245,15 +245,28 @@ public synchronized void shutdownNode() {
...
+            int timeout = Integer.MAX_VALUE; // Default. Wait a very long time.
+            int expired = 0;
+            try {
+                timeout = ConfigParser.parse().getInt("nodeRecycleThreadWaitTimeout");
+            } catch (ConfigParserException e) { // NOSONAR
+                // ignore we have a default already
+            }
             appendMsgToCustomLog("Started NodeRecycleThread to restart the node once all the sessions are completed.");
-            while (getTotalUsed() > 0) {
+            while ((getTotalUsed() > 0) && (expired < timeout)) {
```
sebady: I was about to point at this loop as suspect with any ill behaved tests.   Can we test for a possible hang here with a test?

mach6: Yes. You can create a test that polls the browser and keeps it open indefinitely and run some "well-behaved" tests which reach the `uniqueSessionCount` and trigger the `NodeRecycleThread` ... In this scenario, the `NodeRecycleThread` will keep looping as long as that long running "mis-behaved" test keeps polling..

The purpose of my change is to provide for a configurable behavior which will allow the `NodeRecycleThread` to exit, should something like this happen.

server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
```diff
@@ -224,37 +247,97 @@ public synchronized void shutdownNode() {
...
+
+            try {
+                timeout = ConfigParser.parse().getInt("nodeRecycleThreadWaitTimeout");
+            } catch (ConfigParserException e) { // NOSONAR
+                // ignore we have a default already
+            }
+
+            proxyLogger.fine("Started NodeRecycleThread with " + ((timeout == -1) ? "no" : "a " + timeout + " second")
+                    + " timeout.");
+            while (keepLooping(expired, timeout)) {
```
sebady, line 296: Can we simplify the timeout loop here in run? also would a timeout of 0 be better to indicate "loop forever" (i.e no timeout) instead of "-1"

server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
```diff
@@ -224,37 +247,97 @@ public synchronized void shutdownNode() {
...
+
+        private boolean wasExpired(int expired, int timeout) {
+            return (expired >= timeout) && (timeout != -1);
+        }
+
+        public void shutdown() {
+            LOGGER.entering();
+            running = false;
+            proxyLogger.fine("Interrupting NodeRecycleThread.");
+            interrupt();
```
sebady, line 333: We're setting running to false and interrupting the thread here.  Isn't one sufficient?

server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
```diff
@@ -191,31 +181,64 @@ public TestSession getNewSession(Map<String, Object> requestedCapability) {
...
     private void startNodeRecycleThread() {
         if (!nodeRecycleThread.isAlive()) {
-            appendMsgToCustomLog("Spawning NodeRecycleThread to recycle the node " + getId());
+            proxyLogger.fine("Spawning NodeRecycleThread to recycle the node " + getId());
             nodeRecycleThread.start();
         }
     }
 
+    private void stopNodeRecycleThread() {
+        if (nodeRecycleThread != null) {
```
sebady: Should this test if thread is alive? I don't think  nodeRecycleThread is ever null.

server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
```diff
@@ -224,37 +247,97 @@ public synchronized void shutdownNode() {
...
             try {
                 client.close();
             } catch (IOException e) {
                 LOGGER.log(Level.SEVERE, e.getMessage(), e);
             }
         }
+        stopNodeRecycleThread();
+        // remove the node from the hub registry, if it reported 200 OK for NodeForceRestartServlet request
+        if (responseStatusCode == HttpStatus.SC_OK) {
+            getRegistry().removeIfPresent(this);
```
sebady, line 264: Can we unregister the node in the graceful shutdown path as well?

mach6, line 264: Already do. `requestNodeShutdown` (graceful shutdown) starts the `NodeRecycleThread` which calls `forceNodeShutdown` upon exit and when not interrupted.

server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
mach6: will fix

server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
mach6, line 296: yea. I can do that.

server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
mach6: will fix

mach6, line 333: Actually, I'm going to keep this in place. The reason for setting `running = false` is I check it in the `catch (InterruptedException e)` block of the main `run()` loop.. If `shutdown()` was not called, then something else interrupted the recycle thread, and I want a `SEVERE` message displayed on the hub console.

server/src/main/resources/config/unix/SeLionConfig.json
```diff
@@ -1,6 +1,7 @@
 {
     "restartCycle": 60000,
     "uniqueSessionCount": 50,
+    "nodeRecycleThreadWaitTimeout": 300,
```
mach6, line 4: @sebady I was unsure about this default. Do you think a 5min timeout is a reasonable default?

sebady, line 4: it seemed pretty reasonable to me.

PR #189, opened 2015-10-12
Fixed SeLionRemoteProxy deadlock issue when unregistering the node fr…

sundaramrajendran:
…om Grid if session are still active
- Releasing all the test slots before restarting the node
- Added maximum session reached check before synchronization block to avoid acquiring lock in getNewSession()
- Removed synchronized block in afterSession by updated totalSessionCompleted as volatile

### DIFF
--- a/server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
+++ b/server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
@@ -19,6 +19,8 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
 
 import org.apache.http.HttpResponse;
@@ -33,7 +35,9 @@ import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.message.BasicNameValuePair;
 import org.openqa.grid.common.RegistrationRequest;
 import org.openqa.grid.internal.Registry;
+import org.openqa.grid.internal.SessionTerminationReason;
 import org.openqa.grid.internal.TestSession;
+import org.openqa.grid.internal.TestSlot;
 import org.openqa.grid.selenium.proxy.DefaultRemoteProxy;
 
 import com.paypal.selion.SeLionBuildInfo;
@@ -65,11 +69,12 @@ public class SeLionRemoteProxy extends DefaultRemoteProxy {
     private static final SeLionGridLogger LOGGER = SeLionGridLogger.getLogger(SeLionRemoteProxy.class);
     private SimpleLogger proxyLogger;
     private static final int MAX_SESSION_ALLOWED = 50;
-
-    private int maxSessionsAllowed, totalSessionsCompleted = 0, totalSessionsStarted = 0;
-    private boolean forceShutDown = false;
+    private volatile int totalSessionsCompleted = 0, totalSessionsStarted = 0;
+    private int maxSessionsAllowed;
+    private volatile boolean forceShutDown = false;
     private String machine;
     private NodeRecycleThread nodeRecycleThread = new NodeRecycleThread(getId());
+    private Lock accessLock = new ReentrantLock();
 
     private int getUniqueSessionCount() {
         try {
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
```diff
@@ -163,16 +165,15 @@ public boolean upgradeNode(String downloadJSON) {
     @Override
     public TestSession getNewSession(Map<String, Object> requestedCapability) {
         LOGGER.entering();
-        TestSession session;
+        //verification should be before synchronized block also to avoid unnecessarily acquiring lock
+        if (totalSessionsStarted >= maxSessionsAllowed || forceShutDown) {
+            return logSessionInfo();
+        }
         synchronized (this) {
             if (totalSessionsStarted >= maxSessionsAllowed || forceShutDown) {
```
sebady, line 181: Did you intend to leave this if test here? How do we get here with the above return?

sundaramrajendran, line 181: @sebady Yeah I did. Lets assume totalSessionStarted is 24 and maxSessionAllowed is 25. If two threads enter into getNewSession() both of them pass first if condition. Thread1 acquire the lock, get the new session and increase the count totalSessionStarted to 25, release the lock. Now thread2 enter into synchronized block and get into second if condition. This is rare scenario, but we need this check here to avoid these kind of scenario.

sebady, line 181: That makes sense.  we may want to add a comment somewhere on the 2nd if test.  Also don't we now need to make forceShutDown volatile also if the first test is outside the lock?

sundaramrajendran, line 181: Yeah, I will update both.

server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
```diff
@@ -65,9 +67,9 @@
     private static final SeLionGridLogger LOGGER = SeLionGridLogger.getLogger(SeLionRemoteProxy.class);
     private SimpleLogger proxyLogger;
     private static final int MAX_SESSION_ALLOWED = 50;
-
-    private int maxSessionsAllowed, totalSessionsCompleted = 0, totalSessionsStarted = 0;
-    private boolean forceShutDown = false;
+    private volatile int totalSessionsCompleted = 0;
+    private int maxSessionsAllowed, totalSessionsStarted = 0;
```
mach6: Why don't we also need to make `totalSessionsStarted` `volatile`?

sundaramrajendran: I will update

server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
```diff
@@ -206,13 +217,11 @@ private void stopNodeRecycleThread() {
...
     public void afterSession(TestSession session) {
         LOGGER.entering();
-        synchronized (this) {
-            totalSessionsCompleted++;
-            if (totalSessionsCompleted <= maxSessionsAllowed) {
-                proxyLogger.fine("Completed session #" + totalSessionsCompleted + " (" + session.toString() + ")");
-            }
-            proxyLogger.fine("Total number of slots used: " + getTotalUsed() + " on node: " + getId());
+        totalSessionsCompleted++;
+        if (totalSessionsCompleted <= maxSessionsAllowed) {
```
mach6: This `if` does not seem necessary. `totalSessionsCompleted` is not used to prevent new sessions when `maxSessionsAllowed` is reached in `getNewSession`

server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
```diff
@@ -163,16 +165,17 @@ public boolean upgradeNode(String downloadJSON) {
     @Override
     public TestSession getNewSession(Map<String, Object> requestedCapability) {
         LOGGER.entering();
-        TestSession session;
+        //verification should be before synchronized block also to avoid unnecessarily acquiring lock
+        if (totalSessionsStarted >= maxSessionsAllowed || forceShutDown) {
+            return logSessionInfo();
+        }
         synchronized (this) {
+            //As per double-checked locking pattern need to have check once again
```
mach6, line 179: How about using a `RentrantLock` like we do in `SeLionSauceProxy` instead?

server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
```diff
@@ -186,6 +189,14 @@ public TestSession getNewSession(Map<String, Object> requestedCapability) {
         }
     }
 
+    private TestSession logSessionInfo() {
+        proxyLogger.fine("Was max sessions reached? " + (totalSessionsStarted >= maxSessionsAllowed)
+                + " on node " + getId());
+        proxyLogger.fine("Was this a forcible shutdown? " + (forceShutDown) + " on node " + getId());
```
mach6: Let's re-word this to `a forced shutdown`

sundaramrajendran:
@mach6 Addressed all the review comments

mach6:
@sundaramrajendran @sebady I have put some more changes on top of this PR and opened #190 with the changes. Please review.. closing this PR.

PR #207, opened 2015-11-03
Refactoring for more selion grid servlet html

mach6:
Important note: This PR builds on #202 ..


sebady:
Didn't look at all the js here but the rest of the PR looks good to merge

server/src/main/resources/com/paypal/selion/js/listAllNodes.js
```diff
@@ -0,0 +1,209 @@
...
+
+      // add a <li> item to display
+      function createInfoLi(label, item, valueType, isSelionItem) {
+        // sanitize known values which mean the proxy did not provide any data.
+        function supported(something) {
+          if (proxyClass === 'SeLionRemoteProxy') {
+            return something;
+          }
+          try {
+            if ((something === undefined) || (something === null) || (something === -1)) {
```
sundaramrajendran: We can update this as if (!something || (something === -1)). if (something) will validate false, 0, "", NaN, null, 
undefined.

server/src/main/resources/com/paypal/selion/js/listAllNodes.js
```diff
@@ -0,0 +1,209 @@
...
+              return 'not supported';
+            }
+          } catch (err) {
+            return 'not supported';
+          }
+          return something;
```
sundaramrajendran: We can avoid try block as below:

if (something && (something !== -1)) {
  return something;
}
return 'not supported';

sundaramrajendran:
Overall looks good to me

server/src/main/resources/com/paypal/selion/js/listAllNodes.js
mach6: Thx. I'll make this change.

PR #211, opened 2015-11-09
Fixed coding standard issues

sundaramrajendran:
- Enforces calling super inside constructor
- Prohibitis reassigning values to parameters
- Prohibits if-else when returning boolean
- Prohibitis return of internal arrays
- Enforces never changed fields to final
- Prohibits unnecessary comparisions in boolean expressions
- Enforces literals first in comparisons
- Prohibitis non thread safe singletons
- Import statements allow the use of non-fully qualified names
- Use assertSame/assetNotSame in assertions when testing object referrences equality
- Use asserrTrue/assertFalse to assert boolean values
- Avoid unused import statements

### DIFF
--- a/client/src/main/java/com/paypal/selion/internal/platform/grid/browsercapabilities/SelendroidCapabilitiesBuilder.java
+++ b/client/src/main/java/com/paypal/selion/internal/platform/grid/browsercapabilities/SelendroidCapabilitiesBuilder.java
@@ -36,29 +36,29 @@ class SelendroidCapabilitiesBuilder extends DefaultCapabilitiesBuilder {
     public DesiredCapabilities getCapabilities(DesiredCapabilities capabilities) {
 
         MobileTestSession mobileSession = Grid.getMobileTestSession();
-        capabilities = SelendroidCapabilities.android();
+        DesiredCapabilities tempCapabilities = SelendroidCapabilities.android();
 
         // check if apk exists for native app to set BrowserName to 'selendroid'
         // else set it to 'android'
         if ((new File(mobileSession.getAppLocation()).exists())
                 && ((new File(mobileSession.getAppLocation() + File.separator + mobileSession.getAppName())).exists())) {
-            capabilities.setBrowserName(SELENDROID);
+            tempCapabilities.setBrowserName(SELENDROID);
         } else {
-            capabilities.setBrowserName(ANDROID);
+            tempCapabilities.setBrowserName(ANDROID);
         }
 
-        capabilities.setCapability(MOBILE_NODE_TYPE, mobileSession.getMobileNodeType().getAsString());
-        capabilities.setCapability(SelendroidCapabilities.AUT, mobileSession.getAppName());
-        capabilities.setCapability(SelendroidCapabilities.LOCALE, mobileSession.getAppLocale());
+        tempCapabilities.setCapability(MOBILE_NODE_TYPE, mobileSession.getMobileNodeType().getAsString());
+        tempCapabilities.setCapability(SelendroidCapabilities.AUT, mobileSession.getAppName());
+        tempCapabilities.setCapability(SelendroidCapabilities.LOCALE, mobileSession.getAppLocale());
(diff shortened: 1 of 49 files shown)


client/src/main/java/com/paypal/selion/internal/platform/grid/browsercapabilities/SelendroidCapabilitiesBuilder.java
```diff
@@ -35,30 +35,31 @@
     @Override
     public DesiredCapabilities getCapabilities(DesiredCapabilities capabilities) {
 
+        DesiredCapabilities tempCapabilities = capabilities;
```
mach6: Here the _Object reference_  to `capabilities` is copied. So `capabilities` and `tempCapabilities`, refer to the same object.

Thus, this change is not really fixing the `Avoid reassigning parameters such as 'capabilities'` condition.

In this particular case, the approach should be 

``` java
DesiredCapabilties newCapabilities = SelendroidCapabilities.android(); // Note: SelendroidCapabilities.android() returns a new DesiredCapabilties object. So,now that the original `capabilities` _Object reference_ and values are lost.
newCapabilities.merge(capabilities); // I suspect the fact that we don't merge the pre-existing values is a bug we have today. See `IOSDriverCapabilitiesBuilder` -- it calls `merge()`, for example

// the rest of the changes
```

client/src/main/java/com/paypal/selion/internal/reports/excelreport/DetailsReport.java
```diff
@@ -62,17 +62,18 @@ public DetailsReport(String sReportName) {
 
     int fillData(HSSFSheet sheet, int rowNum, HSSFCellStyle style) {
         logger.entering(new Object[] { sheet, rowNum, style });
+        HSSFCellStyle tempStyle = style;
```
mach6: Problematic: _Object reference_ copy

client/src/main/java/com/paypal/selion/platform/html/support/HtmlElementUtils.java
```diff
@@ -187,40 +187,41 @@ public static By resolveByType(String locator) {
      */
     public static By getFindElementType(String locator) {
         logger.entering(locator);
-        Preconditions.checkArgument(StringUtils.isNotBlank(locator), INVALID_LOCATOR_ERR_MSG);
+        String tempLocator = locator;
```
mach6: Let's give this a better name (like `seleniumLocator`) and create this object after the `Precondition` where the assignment happens.

client/src/main/java/com/paypal/selion/platform/mobile/ios/UIAPicker.java
```diff
@@ -72,10 +72,11 @@ private WebElement findWheelAtIndex(int index) {
     }
 
     private List<String> asList(String allValues) {
+        String tempAllValues = allValues;
```
mach6: Just declare a short-lived variable where the assignment happens, in the `if` loop

server/src/main/java/com/paypal/selion/grid/FileExtractor.java
```diff
@@ -81,6 +81,7 @@ static String getFileNameFromPath(String name) {
         LOGGER.entering(archiveFile);
 
         LOGGER.info("Extracting " + archiveFile);
+        String tempArchiveFile = archiveFile;
```
mach6: Similar feedback. Let's do the assignment to a new variable when/where the assignment happens. Let's keep the 'reads' / conditionals in this method acting on the method parameter `archiveFile`

mach6:
@sundaramrajendran just some minor feedback. otherwise this looks good. thanks for going after the codacy issues!

sundaramrajendran:
Addressed all the review comments

mach6:
updated and pushed to branch https://github.com/paypal/SeLion/tree/sundaramrajendran-raj-coday-fix-errorprone

mach6:
merged to `develop`

PR #222, opened 2015-12-08
Fix #220 - appName and appPath processing bug

### LINKED ISSUE #220: String 'selion-hub-storage:' is not converted when mobile 'appPath' is provided outside the MobileTest annotation.
This logic only happens when the value is passed via the annotation;

https://github.com/paypal/SeLion/blob/develop/client/src/main/java/com/paypal/selion/internal/platform/grid/MobileTestSession.java#L180-L194

mach6:
Processing of appName and appPath in MobileTestSession was not
happening when values were passed as config properties

### DIFF
--- a/client/src/main/java/com/paypal/selion/annotations/MobileTest.java
+++ b/client/src/main/java/com/paypal/selion/annotations/MobileTest.java
@@ -52,35 +52,35 @@ public @interface MobileTest {
     String appName() default "";
 
     /**
-     * Establish the type of device to be used(android, iphone, etc). Default's to <code>iphone</code> Platform version
-     * can be specified as part of the device as:
+     * Establish the type of device to be used (android, iphone, etc). No default. Platform version can be specified as
+     * part of the device as:
      * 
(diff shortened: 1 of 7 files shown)


mach6:
@sebady @sundaramrajendran any concerns with this?

sundaramrajendran:
Looks good to me

sebady:
This specific change looks fine.  I would also recommend fixing the use of this to be consistent.

There are more problems though in this initializeTestSession.
If the configuration for language was specified as "MOBILE_APP_LANGUAGE=fr" the method initializeTestSession will use "en" for lang instead since that is the default value in MobileTest annotation.

mach6:
@sebady The problem you mention is due to the default value in `MobileTest` -- so we need to rethink that ... I don't see any inconsistency in `initializeTestSession` here w.r.t how things work in `WebTest`

mach6:
@sebady @sundaramrajendran I believe I have addressed the review feedback and updated this PR. Please confirm

PR #223, opened 2015-12-08
Code changes to fix SeLion Grid issues

sundaramrajendran:
- Fixed the issue which will allow to read SeLionConfig.json from selionHome directory
- Reduced the probability of node registering automatically to the Grid when SeLionRemoteProxy remove node from Grid for forceshutdown

### DIFF
--- a/server/src/main/java/com/paypal/selion/utils/ConfigParser.java
+++ b/server/src/main/java/com/paypal/selion/utils/ConfigParser.java
@@ -40,7 +40,7 @@ import com.paypal.selion.pojos.SeLionGridConstants;
 public final class ConfigParser {
     private static final SeLionGridLogger LOGGER = SeLionGridLogger.getLogger(ConfigParser.class);
     private static ConfigParser parser = new ConfigParser();
-    private JsonObject configuration;
+    private static JsonObject configuration;
     private static String configFile;
 
     /**
@@ -48,6 +48,14 @@ public final class ConfigParser {
      *         represented by the JSON file passed via the JVM argument <b>SeLionConfig</b>
      */
     public static ConfigParser parse() {
+        LOGGER.entering();
+        if (configuration == null) {
+            try {
+                readConfigFileContents();
+            } catch (IOException e) {
+                throw new ConfigParserException(e);
+            }
+        }
         LOGGER.exiting(parser.toString());
         return parser;
     }
@@ -58,9 +66,13 @@ public final class ConfigParser {
      * @param file
      *            the SeLion Grid config file to use
      */
-    public static void setConfigFile(String file) {
+    public static ConfigParser setConfigFile(String file) {
         LOGGER.entering(file);
-        configFile = file;
+        if (configuration == null) {
+            configFile = file;
+        }
+        LOGGER.exiting(parser.toString());
+        return parser;
     }
 
     /**
@@ -184,20 +196,15 @@ public final class ConfigParser {
     }
 
     private ConfigParser() {
-        try {
-            readConfigFileContents();
-        } catch (IOException e) {
-            throw new ConfigParserException(e);
-        }
-
+        //Intentionally left blank
     }
 
-    private void readConfigFileContents() throws IOException {
+    private static void readConfigFileContents() throws IOException {
         LOGGER.entering();
         InputStream stream = null;
         if (StringUtils.isBlank(configFile)) {
             LOGGER.fine("Config file will be loaded as a resource.");
-            stream = this.getClass().getResourceAsStream(SeLionGridConstants.SELION_CONFIG_FILE_RESOURCE);
+            stream = ConfigParser.class.getResourceAsStream(SeLionGridConstants.SELION_CONFIG_FILE_RESOURCE);
         } else {
             File config = new File(configFile);
             String path = config.getAbsolutePath();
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/paypal/selion/utils/ConfigParser.java
```diff
@@ -48,6 +48,10 @@
      *         represented by the JSON file passed via the JVM argument <b>SeLionConfig</b>
      */
     public static ConfigParser parse() {
+        LOGGER.entering();
+        if (parser == null) {
```
mach6: Codacy Issue found: [Singleton is not thread safe](https://www.codacy.com/app/SeLion/paypal-SeLion/file/1516698859/issues/source?bid=2928425&fileBranchId=3008049#l52)

mach6: @sundaramrajendran Will you please elaborate on the issue this change is fixing? The code change you made here is a revert of a previous change you made. See https://github.com/sundaramrajendran/SeLion/commit/b8064e0686225b38982510fe1c7f0c72748dcb7c#diff-f0886e3373d41f8f4e2ca67ec83de2e4

sundaramrajendran: @mach6 Yeah I have reverted the previous commit. Issue is SeLion not able to read the SeLionConfig.json file from selionHome directory specified from java arguments because of previous change. Since we are creating static instance for ConfigParser, always it will load the SeLionConfig.json from jar file, not the specified file via setConfigFile(String file). Here ConfigParser getting initialized before we are setting config file. So we need lazy initialization to make it work,

mach6: @sundaramrajendran Yes, I see the bug now. Thanks!  This said, I think `ConfigParser` needs a few more changes.
1. We should not call `readConfigFileContents` in any constructor.
2. We should wait to call `readConfigFileContents` on the first invocation of `parse`
3. `setConfigFile` should return the `ConfigParser` instance to make it more like a builder. E.g. - `ConfigParser.setConfigFile(file).parse().getInt(key)`
4. Subsequent calls to `setConfigFile` should not change the static value of `configFile` once `parse` has also been called.

sundaramrajendran:
@mach6 I have addressed all the review comments except adding new constructor. ConfigParser we are using singleton pattern, So I didn't find any importance of adding constructor with configFile as parameter.

mach6:
@sundaramrajendran sorry on the constructor comment.. I noticed my error and actually edited/removed that suggestion in my note.

mach6:
@sundaramrajendran How about?

mach6:
@sundaramrajendran What about these two suggestions? 

> 1. `setConfigFile` should return the `ConfigParser` instance to make it more like a builder. E.g. - `ConfigParser.setConfigFile(file).parse().getInt(key)`
> 2. Subsequent calls to `setConfigFile` should not change the static value of `configFile` once parse has also been called.

server/src/main/java/com/paypal/selion/utils/ConfigParser.java
```diff
@@ -58,9 +66,13 @@ public static ConfigParser parse() {
      * @param file
      *            the SeLion Grid config file to use
      */
-    public static void setConfigFile(String file) {
+    public static ConfigParser setConfigFile(String file) {
         LOGGER.entering(file);
-        configFile = file;
+        if (configuration == null) {
```
mach6: Codacy Issue found: [Singleton is not thread safe](https://www.codacy.com/app/SeLion/paypal-SeLion/file/1536614930/issues/source?bid=2928425&fileBranchId=3008049#l71)

server/src/test/java/com/paypal/selion/utils/ConfigParserTest.java
```diff
@@ -15,26 +15,17 @@
...
 
 import java.io.File;
 
-import org.mockito.Mockito;
-import org.powermock.core.classloader.annotations.PrepareForTest;
-import org.powermock.modules.testng.PowerMockTestCase;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
-@PrepareForTest(ConfigParser.class)
```
mach6, line 29: Doesn't removing this mock object in this test class create a race condition for the rest of the test classes
which may rely on the legit config file?  `{selionHome}/config/SeLionConfig.json` vs `/config/DummySeLionConfig.json`

sundaramrajendran, line 29: It should, But I didn't face the issue. Let me check once again and update the code if required.

server/src/main/java/com/paypal/selion/utils/ConfigParser.java
mach6: Codacy Issue found: [Singleton is not thread safe](https://www.codacy.com/app/SeLion/paypal-SeLion/file/1571905575/issues/source?bid=2928425&fileBranchId=3008049#l71)

sundaramrajendran:
I have fixed the ConfigParserTest with PowerMockito.

server/src/main/java/com/paypal/selion/utils/ConfigParser.java
mach6, line 71: Codacy Issue found: [Singleton is not thread safe](https://www.codacy.com/app/SeLion/paypal-SeLion/file/1595242861/issues/source?bid=2928425&fileBranchId=3008049#l71)

PR #230, opened 2016-01-06
Fix #226, #229 and misc cleanup

### LINKED ISSUE #226: UiObject.setText() calls UiObject.clearText(), but SeLionAppiumAndroidDriver's setText already uses its cleartext()
Since SeLionAppiumAndroidDriver's setText already uses cleartext it is redundant to call UiObject.clearText() from UiObject.setText().

### LINKED ISSUE #229: Unable to create android element of type UiTextView
We are getting an error when trying to generate Android page objects that contain "UiTextView" elements.  It appears the code generator is looking for an element with the name "UiTextField" instead of "UiTextView":  https://github.com/paypal/SeLion/blob/develop/codegen/src/main/java/com/paypal/selion/elements/AndroidSeLionElementList.java

mach6:
- Fix #226 redundant call to clear text
- Fix #229 UiTextView elements are not supported by default via code-generator
- Cleanup print config messages
- Remove some redundant initializers

### DIFF
--- a/client/src/main/java/com/paypal/selion/configuration/Config.java
+++ b/client/src/main/java/com/paypal/selion/configuration/Config.java
@@ -1,5 +1,5 @@
 /*-------------------------------------------------------------------------------------------------------------------*\
-|  Copyright (C) 2014-15 PayPal                                                                                       |
+|  Copyright (C) 2014-2016 PayPal                                                                                     |
 |                                                                                                                     |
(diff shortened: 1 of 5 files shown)


mach6:
@sundaramrajendran please review

sundaramrajendran:
@mach6 Looks good to me. Just update copyright year for AndroidSeLionElementList

PR #238, opened 2016-01-20
Fix exception handling when SauceLabs goes down

sebady:
Getting the number of Test cases running from sauceLabs fails
we return an empty string (after several retries to Sauce)
which causes an illegalStateException on parsing in
SeLionSauceProxy. Now log and continue.


mach6:
@sebady as soon as the CI succeeds, go ahead and merge this please.

PR #241, opened 2016-01-25
Add WerFault process on Windows

sebady:
Add WerFault.exe (Windows Error Reporting) process to the list of process
we search for and kill on a node recycle.


mach6:
@sebady okay with me. please merge. 

note: I suspect if a werfault happens during the node shutdown (browser cleanup), it could be missed -- since we search by pid -- once restarted the node would have a new pid

sebady:
We also are searching for child processes during node cleanup and I'm pretty certain  werfault.exe is not a child of the node process.  So there is more work to do here.

mach6:
@sebady :+1: LGTM. Please merge this.

PR #249, opened 2016-02-22
Add before and after screenshot events to the element event listener.

renelux:
We need this method to make changes to the webpage before a screenshot gets taken. SeLion by default takes screenshots of the whole page (which we love). But a lot of websites nowadays are using floating headers/elements, this means that when the page is scrolled the header shows in the middle of the page. 

This becomes a problem for the QA's that review the screenshots, with this change we can add a scroll up before the screenshot gets take, and scroll down afterwards.

### DIFF
--- a/client/src/main/java/com/paypal/selion/platform/html/AbstractElement.java
+++ b/client/src/main/java/com/paypal/selion/platform/html/AbstractElement.java
@@ -387,6 +387,8 @@ public abstract class AbstractElement implements Clickable, Hoverable {
         logger.entering();
         processAlerts(Grid.getWebTestSession().getBrowser());
 
+        dispatcher.beforeScreenshot(this);
+        
         String title = "Default Title";
         try {
             title = Grid.driver().getTitle();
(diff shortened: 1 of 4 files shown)


mach6:
@renelux will take a look at this soon. sorry I missed you on gitter

renelux:
@mach6 no worries not in a rush. Issues is already identified just need to find a good way to fix it. This way would work (already tried it locally) you just need to approve it. No pressure ;)

mach6:
@renelux I'm okay with this change.I'll merge it in the morning.

mach6:
@renelux sorry, I just noticed. I need you to fix the author (~/.gitconfig `user.email` and `user.name`) of this commit so that it complies the project's contribution policy.

renelux:
@mach6 I updated my commit info

mach6:
thx @renelux .. you sure about the email address? this will be the 3rd address the history has for you. :)

renelux:
@mach6 sometimes I work on my PayPal machine and sometimes on Symbio's that is why it is not always the same :) The paypalcorp.com are the new address I think, but @paypal.com still works so I can change it for you.

renelux:
@mach6 I updated my email again.

mach6:
this is merged.

PR #268, opened 2016-03-31
Bypass Grid#getRemoteNodeInfo for sauce

mach6:


mach6:
Copyright headers need to be updated before this is merged.

PR #271, opened 2016-04-06
update phantomjs to v2.1.1 in server

sebady:


mach6:
I think this looks good @sebady 
One thing to note is our travis ci build will still use 1.9.8 -- the version they have installed.

PR #277, opened 2016-04-22
Move two SeLionConfig.json params to Selenium config

mach6:
The following two parameters are moved out of
SeLionConfig.json and into the hubconfig.json
and/or nodeconfig.json locations, where they
should be. This change allows these values to be
set globally by the hub or per node, which
will override any value specified by the hub.
1. uniqueSessionCount
2. nodeRecycleThreadWaitTimeout

### DIFF
--- a/server/src/main/resources/config/windows/nodeConfig.json
+++ b/server/src/main/resources/config/windows/nodeConfig.json
@@ -29,6 +29,7 @@
         "port": 5555,
         "host": "ip",
         "register": true,
-        "hubPort": 4444
+        "hubPort": 4444,
+        "nodeRecycleThreadWaitTimeout": 300
     }
 }
--- a/server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
+++ b/server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
@@ -58,7 +58,6 @@ import com.paypal.selion.node.servlets.NodeAutoUpgradeServlet;
 import com.paypal.selion.node.servlets.NodeForceRestartServlet;
 import com.paypal.selion.pojos.BrowserInformationCache;
 import com.paypal.selion.pojos.SeLionGridConstants;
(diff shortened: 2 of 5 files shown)


mach6:
Will you please update the javadocs for `ConfigProperty#SELENIUN_PORT` to include this text `Used when {@link ConfigProperty#SELENIUM_RUN_LOCALLY} is <b>false</b>` which is also called out for `ConfigProperty#SELENIUM_HUB`? If you can update the `gh-pages` branch for the same, I'd appreciate it.

server/src/main/resources/config/windows/nodeConfig.json
```diff
@@ -29,6 +29,7 @@
         "port": 5555,
         "host": "ip",
         "register": true,
-        "hubPort": 4444
+        "hubPort": 4444,
+        "nodeRecycleThreadWaitTimeout": 300
```
sebady, line 33: Since these values are relevant only to SeLion (and not selenium), I think the SeLionConfig.json file is preferred here.
I'm not seeing the advantage to have this in nodeConfig.

mach6, line 33: Well, the point is Selenium provides a path for hub and node configuration and a registry to keep track of it.. There's nothing about the registry config implementation that is strictly selenium-core specific.

One key advantage to this approach is the hub can set value X and each node can override it with value Y.. The existing `SeLionConfig.json` approach does not support this -- We'd have to change the node boot code to include the override with the registration request --- which at that point is going to put the new value in the exact same config (selenium) source.

mach6, line 33: Building on my previous note.. Configuration for SeLion servlets such as `TransferServlet` or process handlers such as `WindowsProcessHandler` wouldn't make sense to put in the selenium hub configuration -- these things don't read from / have access to the same registry, for one.. Thus, `SeLionConfig.json` and/or command-line dash options processed by SeLion-Grid is more suitable for their configuration.

Issue #286, opened 2016-05-10
Code-generator does not allow you to place *Ext.java classes in src/test/java


mach6:
If a person want's to limit their `{PageObject}Ext.java` classes to the `src/test/java` (test) scope, the code generator will not consider them. 

The code-generator plugin should allow for PageObject classes to be limited to 'test' only code.

`CodeGeneratorMojo#sourceDir()` is the root of the issue here.

mach6:
fix merged with #287

Issue #285, opened 2016-05-10
Code-generator always outputs to target/generated-sources


mach6:
if you bind code-generator to the `generate-test-sources` phase it should output to the `target/generated-test-sources` directory. In addition, the plugin should allow the user to override the output location.

Steps to reproduce: Use this xml in the project pom.

``` xml
<plugin>
  <groupId>com.paypal.selion</groupId>
  <artifactId>SeLion-Code-Generator</artifactId>
  <version>1.0.0</version>
  <executions>
    <execution>
       <phase>generate-test-sources</phase>
       <goals>
         <goal>generate</goal>
       </goals>
    </execution>
  </executions>
</plugin>
```

mach6:
fix merged with #287

PR #287, opened 2016-05-11
Fix #284, #285, #286 - better support for 'generate-test-sources'

### LINKED ISSUE #284: Code-generator does not allow you to place PageYAML in test resources path
There is a bug where the code-generator expects to pick up PageYAML in `src/main/resources`. This is considered a bug as it will result in all PageYAML files (unless they are filtered out) ending up in the non "test" package (jar/war) -- which a user may or may not want to do.

mach6:
- Updated maven plugin to require maven 3
- Added ability to override outputDirectory of generated code
- Fixed support for 'generate-test-sources' phase
- Added new 'generate-test' goal
- Fixed ability to read PageYAML from /src/test/resources
- Fixed ability to use *Ext.java which reside in /src/test/java

### DIFF
--- a/codegen/src/main/java/com/paypal/selion/plugins/AbstractBaseCodeGeneratorMojo.java
+++ b/codegen/src/main/java/com/paypal/selion/plugins/AbstractBaseCodeGeneratorMojo.java
@@ -1,0 +1,294 @@
+/*-------------------------------------------------------------------------------------------------------------------*\
+|  Copyright (C) 2016 PayPal                                                                                          |
+|                                                                                                                     |
+|  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance     |
+|  with the License.                                                                                                  |
+|                                                                                                                     |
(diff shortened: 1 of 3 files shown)


mach6:
@sebady please review

codegen/src/main/java/com/paypal/selion/plugins/AbstractBaseCodeGeneratorMojo.java
```diff
@@ -0,0 +1,294 @@
...
+     */
+    abstract String generatedSourcesDir();
+
+    /**
+     * @return the resources directory which contains the {@link #baseFolder}
+     */
+    abstract String resourcesDir();
+
+    private void displayProjectInformation() {
+        Log logger = Logger.getLogger();
```
sebady, line 111: Shouldn't we use the base class getLog() here to get the plugin logger?

sebady, line 111: I see this was in the original code here.

PR #290, opened 2016-05-12
Change archetype to use test scope

### LINKED ISSUE #289: SeLion archetype should use test scope
Since SeLion is a test framework, the archetype new project `pom.xml` should default to `<scope>test</scope>` for SeLion.jar dependency and code generation.

mach6:
- Closes #289
- All test code and assists are included in the archetype are
  now properly associated with the test scope.


archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -34,36 +34,35 @@
             <groupId>com.paypal.selion</groupId>
             <artifactId>SeLion</artifactId>
             <version>${selion.version}</version>
+            <scope>test</scope>
```
mach6: Perhaps I should leave this as it was for this artifact. `TestServerUtils` and `AppLogger` are the `application` under test for the purposes of the included sample code. The problem is they depend on downstream dependencies of the SeLion artifact -- requiring it (the SeLion artifact) to be in `compile` scope. Ideally, I'd like to convey that SeLion should be in `test` scope when used solely for writing/running tests. It should be in the `compile` scope only when creating a re-usable artifact that adds to / sits on-top of a SeLion one.

I suppose another option is to explicitly define the transitive `<dependencies>` of `TestServerUtils` and `AppLogger` in the `compile` scope.

@sebady your thoughts here?

sebady: I would vote we keep things simple in the archetype pom file.  Changing to test scope for SeLion is nice to have but not at the cost of making the pom dependencies overly complicated.

mach6: yea. I think I'll revert the change here.

PR #295, opened 2016-05-27
SeLionRemoteProxy shutdown fixes

mach6:
1. Make the proxy unregister from the hub
2. Handle errors when the shutdown request
   fails for whatever reason
3. Handle situation where proxy registraion
   includes a uniqueSessionCount but proxy
   does not supprt canForceShutdown
4. Misc proxyLogger message improvements

### DIFF
--- a/server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
+++ b/server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
@@ -41,6 +41,7 @@ import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.message.BasicNameValuePair;
 import org.openqa.grid.common.RegistrationRequest;
+import org.openqa.grid.common.exception.RemoteUnregisterException;
 import org.openqa.grid.internal.Registry;
 import org.openqa.grid.internal.SessionTerminationReason;
 import org.openqa.grid.internal.TestSession;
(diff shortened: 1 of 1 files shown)


mach6:
@sebady please review

server/src/main/java/com/paypal/selion/proxy/SeLionRemoteProxy.java
```diff
@@ -444,22 +451,35 @@ public synchronized void forceNodeShutdown() {
         HttpResponse response = sendToNodeServlet(NodeForceRestartServlet.class, nvps);
         if (response == null) {
             proxyLogger.warning("Node " + getId() + " failed to shutdown and returned a null response.");
-            LOGGER.exiting(false);
+            // stop the polling thread, mark this proxy as down, force this proxy to re-register
+            teardown("it failed to shutdown and must re-register");
```
mach6, line 455: wasn't sure which was a better approach here.. 
1. This (as I have done here)
2. Increase `uniqueSessionCount` by `uniqueSessionCount + uniqueSessionCount` and let this proxy continue
3. Track the failure as a new member variable which can be queried or displayed .. something like `hasErrors`
4. 2 & 3
5. Other....

sebady, line 455: "1" seems the most reasonable to me.

If the node is "healthy" (it did shutdown despite response) it will re-register and cause a new proxy instance to be created. If it's not healthy (didn't really shut down and perhaps is hung or whatever) then we don't want it to be be registered with hub and have it's proxy lying around in a "shutting down" state.

PR #298, opened 2016-06-14
Fix a couple codacy reported issues in dataprovider component

mach6:
- Unused import
- mismatch @param in javadoc
- method in USER.java test class
- Missing default in switch
- Default and empty constructor w/o comment


mach6:
looking into the test failure now

PR #302, opened 2016-06-21
Fix #301 - don't log the browser to info every time click occurs

### LINKED ISSUE #301: Everytime a button or link is clicked, INFO is printed out to console with the browser
Everytime a button or link is clicked, INFO is printed out to console with the browser.  Its a bit noisy to constantly print out the browser type every time an element is clicked.

ILikeToNguyen:
Fix #301 - don't log the browser to info every time click occurs.  Log that information to finest.

### DIFF
--- a/client/src/main/java/com/paypal/selion/platform/html/AbstractElement.java
+++ b/client/src/main/java/com/paypal/selion/platform/html/AbstractElement.java
@@ -435,7 +435,7 @@ public abstract class AbstractElement implements Clickable, Hoverable {
 
     protected void validatePresenceOfAlert() {
         String browser = Grid.getWebTestSession().getBrowser();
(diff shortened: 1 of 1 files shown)


client/src/main/java/com/paypal/selion/platform/html/AbstractElement.java
```diff
@@ -435,7 +435,7 @@ private boolean doesNotHandleAlerts(String browserFlavor) {
 
     protected void validatePresenceOfAlert() {
         String browser = Grid.getWebTestSession().getBrowser();
-        logger.info(browser);
+        logger.finest("Validating presense of alert with browser " + browser);
```
mach6: spelling error.

PR #303, opened 2016-06-27
Add initial geckodriver /  marionette support

sebady:
- Add iniitial support for using geckodriver with Firefox browser.
  (Selenium 2.53 does not work with Firefox 47.)
- Add a GeckoCapabilitiesBuilder for geckodriver.
- Add client config option for enabling Geckodriverand
  specifying Geckodriver path for local runs.
- Add marionette support in Grid server.
- Also added a test security manager to prevent System.exit()
  from some node servelet tests.


mach6:
:+1:  @sebady please go ahead and merge this to `develop`, if you are comfortable with it.

sebady:
I really need to look at this a bit more against the latest marionette support in selenium.

PR #304, opened 2016-06-28
Make Config ThreadSafe to allow overwriting non-global property from code.

renelux:
This is change gives a user the ability to overwrite certain properties from the code without it affecting other runs.

This is beneficial in the following situation:

``` java
//This would affect all the running test cases by overwriting the UserAgent to iPhone.
Config.setConfigProperty(ConfigProperty.SELENIUM_USERAGENT, "Mozilla/5.0 (iPhone; CPU iPhone OS 5_0 like Mac OS X) AppleWebKit/534.46 (KHTML, like Gecko) Version/5.1 Mobile/9A334 Safari/7534.48.3");

//Now we can use the following which only affects the test case that we are executing.
Config.setConfigPropertyThreadSafe(ConfigProperty.SELENIUM_USERAGENT, "Mozilla/5.0 (iPhone; CPU iPhone OS 5_0 like Mac OS X) AppleWebKit/534.46 (KHTML, like Gecko) Version/5.1 Mobile/9A334 Safari/7534.48.3");
```

Or in our situation where the amount of cases is so big that it is unmanageable to manage all in suite files with LocalConfig properties.

The change should be fully backwards compatible.

### DIFF
--- a/client/src/main/java/com/paypal/selion/configuration/ConfigManager.java
+++ b/client/src/main/java/com/paypal/selion/configuration/ConfigManager.java
@@ -54,35 +54,31 @@ import com.paypal.selion.logger.SeLionLogger;
 public final class ConfigManager {
 
     // Map for each TestNG test name -> local SeLion config associated with <test>
-    private static Map<String, LocalConfig> configsMap = new ConcurrentHashMap<String, LocalConfig>();
+    private static Map<Thread, LocalConfig> configsMap = new ConcurrentHashMap<Thread, LocalConfig>();
 
     private ConfigManager() {
         // Utility class. So hide the constructor
     }
 
     /**
-     * Adds the local configuration {@link LocalConfig} associated with name. Over-rides any config with the same name.
+     * Adds the local configuration {@link LocalConfig} associated with thread. Over-rides any config with the same thread.
      * 
      * @param config
      *            The LocalConfig.
-     * @param name
-     *            The name to associate with local config.
      */
-    public static synchronized void addConfig(String name, LocalConfig config) {
-        SeLionLogger.getLogger().entering(new Object[] { name, config });
-        checkArgument(StringUtils.isNotBlank(name),
-                "A testname for which configuration is being added cannot be null (or) empty.");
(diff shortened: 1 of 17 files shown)


mach6:
@renelux Thanks for this. I'm confused though. Why wouldn't you update the test java code to use `LocalConfig` to meet your needs? Either way, you'd need to make java code changes.

renelux:
@mach6 I updated the PR to use enhance LocalConfig. I wasn't able to run all test locally so I will have to see if nothing breaks.

The LocalConfig was bound to the TestNG testname which is not unique for each case. So even if using LocalConfig it would overwrite all settings based on the testname that a case is in.

client/src/main/java/com/paypal/selion/configuration/ConfigManager.java
```diff
@@ -54,35 +54,31 @@
...
      * @param config
      *            The LocalConfig.
-     * @param name
-     *            The name to associate with local config.
      */
-    public static synchronized void addConfig(String name, LocalConfig config) {
-        SeLionLogger.getLogger().entering(new Object[] { name, config });
-        checkArgument(StringUtils.isNotBlank(name),
-                "A testname for which configuration is being added cannot be null (or) empty.");
+    public static synchronized void addConfig(LocalConfig config) {
```
sebady, line 69: @renelux: This change will break clients.  
My recollection is test names were unique in suite files. Has that changed ?

Also the changes here will not work in a parallel suite like this one:

```
<suite thread-count="2" name="SeLion Tests Suite" parallel="classes">
    <parameter name="browser" value="*firefox" />
    <test name="Parallel1">
        <parameter name="browser" value="*chrome" />
        <classes>
            <class name="com.paypal.selion.configuration.A" />
        </classes>
    </test>
</suite> 
```

The class A will get firefox and not chrome browsers.

sebady:
@renelux I'm closing this pull request. If you still want to follow up on these changes please re-open this PR or open a github issue for the limitations you are experiencing with configuration of SeLion.

PR #329, opened 2016-10-04
Fix #328 and re-work additional capabilities support

### LINKED ISSUE #328: Not able to add more than one customCapabilities provider.
### SeLion Version

1.1.0
### Component

SeLion-Client
### Expected Behavior

Selion provides option to add one or more custom capability provider to set capabilities for the driver instance to be created. If more than one custom capability provider is present they should be separated by commas 
### Actual Behavior

When more than once custome Capability provider is passed with classnames separated by commas, only the first class is used. The rest are ignored
### Steps to Reproduce

Create two custome capabilty providers, (classes that extend DefaultCapabilitiesBuilder) Pass the classnames separated by commas as the value of the SELENIUM_CUSTOM_CAPABILITIES_PROVIDER property.

mach6:

---

Fix support for specifying more than one capabilities
provider, element event listener, or event listener via
configuration as comma separated values.
Also;
- Add ability to specify `additionalCapabilitiesBuilders`
  through the `WebTest` and `MobileTest` annotations
- Change `customCapabilities` to a `<test>` local config
  property
- Delete `UserDefinedCapabilitiesBuilder` -- this build
  was loaded through the ServiceLoader mechanism and
  existed purely to read the `additionalCapabilities`
  which a user specified via the `@WebTest` or `@MobileTest`
  annotation. Since support for reading these capabilities
  from the annotation is critical path, it didn't make
  sense (to me) that we had this logic in a `ServiceLoader`
  which could be disabled. The logic to do the initialization
  is now in `UserCapabilitiesBuilder` -- which is NOT loaded
  through the `ServiceLoader` and/or configuration
- Refactor `additionalCapabilities` logic in `AbstractTestSession`
  to the class `CapabilitiesHelper`
- Update `Config` and `LocalConfig` to support `List<Object>` return
  types which could be coming from commons-config
- Update tests
- Added our copyright to a file which was missing it.
- `@Deprecate` `ExtendedConfig` -- it has been a discouraged approach
  for defining additional `DesiredCapabilited` via `@BeforeMethod` for
  some time.


mach6:
@sebady @ILikeToNguyen please review this. Note there are test failures with the default suite due to the recent testng version change. If you revert back to the old testng version, the failures go away.

client/src/main/java/com/paypal/selion/internal/platform/grid/browsercapabilities/CapabilitiesHelper.java
```diff
@@ -56,35 +63,140 @@ private CapabilitiesHelper() {
...
+            } catch (InstantiationException | IllegalAccessException e) {
+                throw new IllegalStateException("Unable to apply desired capabilities from " + builder.getName(), e);
+            }
+        }
+
+        logger.exiting(caps);
+        return caps;
+    }
+
+    private static final Map<String, Object> parseIntoCapabilities(String[] capabilities) {
```
mach6: Will fix

client/src/test/java/com/paypal/selion/platform/grid/WebTestTest.java
```diff
@@ -44,33 +63,66 @@ public void testUnamedSessionMultipleThreadsAndInvocations() {
...
+        assertEquals(args.get(0).getAsString(), "--bogus-arg");
+    }
+
     @Test(testName = "testCapabilityViaTestResult", groups = "functional")
     @WebTest
+    @Deprecated
     public void testCapabilityViaTestResult() {
-        assertEquals(Grid.getWebTestSession().getAdditionalCapabilities().getCapability("useBooleanDCCaps"), Boolean.TRUE);
+        // @deprecated assertions
+        assertEquals(Grid.getWebTestSession().getAdditionalCapabilities().getCapability("useBooleanDCCaps"),
```
mach6: Will fix

client/src/main/java/com/paypal/selion/configuration/Config.java
```diff
@@ -259,78 +260,122 @@ public synchronized static void initConfig(Map<ConfigProperty, String> initialVa
...
+     *
      * @param configProperty
      *            The configuration element to set
      * @param configPropertyValue
      *            The value of the configuration element
      * @throws IllegalArgumentException
      *             If problems occur during the set
      */
-    public static synchronized void setConfigProperty(ConfigProperty configProperty, String configPropertyValue) {
+    public static synchronized void setConfigProperty(ConfigProperty configProperty, Object configPropertyValue) {
```
sebady, line 375: Since this is a public API (on a minor release) it would be preferable to deprecate the old method and add the new one.

client/src/main/java/com/paypal/selion/internal/platform/grid/browsercapabilities/CapabilitiesHelper.java
```diff
@@ -56,35 +63,140 @@ private CapabilitiesHelper() {
...
+    }
+
+    /**
+     * Acquire capabilities from the TestNG {@link InvokedMethodInformation}
+     * 
+     * @param methodInfo
+     *            the TestNG {@link InvokedMethodInformation}
+     * @return the provided {@link DesiredCapabilities} which are associated with the {@link InvokedMethodInformation}
+     */
+    @Deprecated
```
sebady, line 134: Why is this marked as deprecated?

client/src/main/java/com/paypal/selion/annotations/WebTest.java
```diff
@@ -49,42 +51,48 @@
...
      * }
      * </pre>
      */
     String[] additionalCapabilities() default {};
 
     /**
+     * Additional {@link DefaultCapabilitiesBuilder}s that this web test should use. Capabilities which are
+     * constructed this way are merged with the results of any/all globally applicable capability builders.
+     */
+    Class<? extends DefaultCapabilitiesBuilder>[] additionalCapabilitiesBuilders() default {};
```
sebady, line 69: Is an array of builders here really required? Isn't supporting just one capability builder sufficient per web/mobiletest?

client/src/main/java/com/paypal/selion/internal/platform/grid/browsercapabilities/CapabilitiesHelper.java
mach6: Because `ExtendedConfig` (which this method uses) is getting `@Deprecated` with this change set. Once removed, this method will also be removed. I want to EOL the approach as an option for specifying additional desired capabilities

client/src/main/java/com/paypal/selion/configuration/Config.java
mach6: The change does not break compilation and it (the method implementation) is already calling `public void setProperty(String key, Object value)` in commons-config.  Thus, there is no need to deprecate and create a new method.

client/src/main/java/com/paypal/selion/annotations/WebTest.java
mach6: in most cases, no it is not required --  So this works;

``` java
@WebTest( additionalCapabilitiesBuilders = MyClass.class )
```

However, if a user wants to have some sort of separation of concerns, the array here can help.  

For example, let's say a user that wants their own SauceCapabilitiesBuilder and ChromeCapabilitiesBuilder.. Furthermore, they only want their SauceCapabilitiesBuilder to do something when SeLion is configured to use sauce labs or the ChromeCapabiltiesBuilder to do something when SeLion is configured to use chrome. In this example, they'd could not utilize a single `@WebTest` annotated test method and have it work as expected, based on runtime configuration, unless we had array support.

``` java
@WebTest( additionalCapabilitiesBuilders = { MySauceCapabilitiesBuilder.class, MyChromeCapabilitiesBuilder.class } )
```
- `-DSELION_USE_SAUCE_LABS=true` could cause the `MySauceCapabilitiesBuilder` to apply some capabilities
- `-DSELION_SELENIUM_BROWSER=chrome` could cause the `MyChromeCapabilitiesBuilder` to apply some capabilities

PR #331, opened 2016-10-11
Add brower build targets to travis

sebady:

---

Add additional build targets to travis CI to also run our
chrome and firefox suites locally.
Also moved the "deploy" step to phantom target.
Also reduced the number of threads in some suites
Also added support for passing chrome options in Config:
Add "--no-sandbox" for chrome execution to work around
chrome on OpenVZ.


.travis.yml
```diff
@@ -1,14 +1,13 @@
 language: java
-sudo: false
+sudo: required
```
mach6: hmm.. this change means we won't build using container based infra .. and thus slower builds :( ..   https://docs.travis-ci.com/user/migrating-from-legacy/

client/src/main/java/com/paypal/selion/configuration/Config.java
```diff
@@ -373,6 +373,18 @@ public static synchronized void setConfigProperty(ConfigProperty configProperty,
...
+         * Adds additional chromedriver chromeOptions args to be used with local/remote runs using the chrome
+         * browser. <br>
+         * The string value here is specified as a comma separated list of arguments. Arguments with
+         * associated value should be separated by a '=' sign.
+         *
+         * For e.g.
+         *
+         * "start-maximized, user-data-dir=/some/path"
+         */
+        SELENIUM_CHROME_OPTIONS("chromeOptions", "", false),
```
mach6: I don't know about this. My pending PR gives another path.

mach6: Also anything comma separated will automatically be turned into a `List` by commons-config and calling `getConfigProperty` will only return the first value.

client/src/main/java/com/paypal/selion/internal/platform/grid/browsercapabilities/ChromeCapabilitiesBuilder.java
```diff
@@ -58,4 +67,12 @@ private String getBinaryPath() {
                 Config.getConfigProperty(ConfigProperty.SELENIUM_CHROMEDRIVER_PATH));
         return location;
     }
+
+    /*
+     * Returns the comma separated chromedriver chromeoptions as a List of Strings after white space removal.
+     */
+    private List<String> getChromeOptions() {
+        String chromeOptions = Config.getConfigProperty(ConfigProperty.SELENIUM_CHROME_OPTIONS);
+        return Arrays.asList(chromeOptions.split("\\s*,\\s*"));
```
mach6: As coded, this will never get a string with `commas` in it. The call to `getConfigProperty()` will always return just the first item.

.travis.yml
```diff
@@ -22,3 +21,60 @@ notifications:
...
+        # See https://sites.google.com/a/chromium.org/chromedriver/capabilities#TOC-Using-a-Chrome-executable-in-a-non-standard-location
+        sudo ln -sf $PWD/chrome-linux/chrome-wrapper /usr/bin/google-chrome
+        export CHROMEDRIVER_VERSION=`curl -s http://chromedriver.storage.googleapis.com/LATEST_RELEASE`
+        curl -L -O "http://chromedriver.storage.googleapis.com/${CHROMEDRIVER_VERSION}/chromedriver_linux64.zip"
+        unzip chromedriver_linux64.zip && chmod +x chromedriver && sudo mv chromedriver /usr/local/bin
+        google-chrome --version
+        mvn test -pl client -DsuiteXmlFile=Chrome-Suite.xml -DSELION_SELENIUM_RUN_LOCALLY=true \
+          -DSELION_SELENIUM_CHROMEDRIVER_PATH=/usr/local/bin/chromedriver \
+          -DSELION_SELENIUM_CHROME_OPTIONS="--no-sandbox" -B -V
+      fi
```
mach6, line 50: Is this approach better than running against Sauce? It seems to come at a cost..

.travis.yml
sebady: I couldn't get our travis builds working w/o sudo.  One of the issues was how chromedriver finds the chrome binary location.  I could revisit the "no-sudo" build with some additional change to support specifying the chrome binary location.

sebady: actually I also recall that chrome is also not supported any longer on _precise_ the ubuntu version that the container base infra is using.  So if fast builds are important we should use a remote chrome on sauce or live with slower builds on a new ubuntu version.

client/src/main/java/com/paypal/selion/internal/platform/grid/browsercapabilities/ChromeCapabilitiesBuilder.java
mach6: I should have noted, my pending PR will include `getListConfigProperty` to get the `List` back which commons-config automatically (and by default) creates by parsing the comma separated values (internally)

client/src/main/java/com/paypal/selion/configuration/Config.java
sebady: We could disable this commons config behavior and Config clients choose how to parse the property values into lists.

.travis.yml
sebady: Yes.  It certainly comes at additional cost.   I think we should really build with both because a local execution with runlocally enabled is a common use case for users of SeLion client.
Also sauce connect has some limitations with PR builds.  See  https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions
and might be a tad bit slower.  
I'd like to add sauce as another matrix'd build and compare the build times then we can take one path or the other.

client/src/main/java/com/paypal/selion/configuration/Config.java
mach6: Yes, we can disable it or change the delimiter. I'd prefer not to.

.travis.yml
mach6: sounds like a good tact.. I'd like to see the comparisons too

mach6: Also, fyi yes on the limitations. In the current .travis.yml this is already handled. Let's talk offline, I'll share the details.

sebady:
I have updated this PR to not depend on sudo to allow us to continue using travis container based builds.

mach6 (approved):
LGTM. I like this!

sebady:
One additional improvement could be to move all these additional browser/driver files out of the root into the "target" folder?

mach6:
> One additional improvement could be to move all these additional browser/driver files out of the root into the "target" folder?

sounds like a good idea.. if it is not a trivial change, I'd suggest  -- not super important on a throw away travis instance. :)

PR #334, opened 2016-10-14
Fix session sharing bugs

mach6:

---
- Allow session sharing @Test methods to be `enabled = false`
- Throw an exception for undefined or unclear test priorities
- Add additional tests for session sharing and update test suites
- Also in this change set;
  - Change the point at which an exception is thrown from
    SeleniumGridListener#beforeInvocation -- deprecate the method
    Grid#getThreadLocalException) as a result of this change

### DIFF
--- a/client/src/main/java/com/paypal/selion/internal/platform/grid/SeleniumGridListener.java
+++ b/client/src/main/java/com/paypal/selion/internal/platform/grid/SeleniumGridListener.java
@@ -18,6 +18,7 @@ package com.paypal.selion.internal.platform.grid;
 import java.io.File;
 import java.lang.management.ManagementFactory;
 import java.lang.management.RuntimeMXBean;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
@@ -107,10 +108,10 @@ public class SeleniumGridListener implements IInvokedMethodListener, ISuiteListe
 
                 // For session sharing, we need to ensure @Test methods are priority based.
                 if (method.isTestMethod()) {
+                    // we need to ensure each @Test method is annotated
                     if (isLowPriority(method)) {
                         // For session sharing tests, Need to create new session only for Test (Web or Mobile) with
-                        // highest
-                        // priority (first test, smallest number) in the class.
+                        // highest priority (first test, smallest number) in the class.
                         testSessionSharingRules(method);
                     } else {
                         return;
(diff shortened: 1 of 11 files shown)


client/src/main/java/com/paypal/selion/internal/platform/grid/SeleniumGridListener.java
```diff
@@ -198,15 +207,26 @@ private boolean isLowPriority(IInvokedMethod method) {
...
 
     private boolean isHighPriority(IInvokedMethod method) {
+        try {
+            checkForTestAnnotation(method);
+        } catch (IllegalStateException e) {
+            //ignore the exception, we will already have this exception from beforeInvocation
+            //when it calls isLowPriority. Instead we need to assume true to allow afterInvocation
+            //to fall into the portion where it closes the session
+            return true;
+        }
```
sebady, line 215: Should we consider two distinct private methods here?  One that throws and one that returns a boolean if the annotation check fails.

sebady:
LGTM.  Just one minor comment.

mach6:
@sebady I don't see any comment. :)

client/src/main/java/com/paypal/selion/platform/grid/Grid.java
```diff
@@ -93,17 +98,10 @@ public static long getExecutionTimeoutValue() {
...
-            if (exception instanceof RuntimeException) {
-                throw (RuntimeException) exception;
-            }
-            // convert the checked exception into a runtime exception.
-            throw new RuntimeException(exception.getMessage(), exception);
-        }
-
         AbstractTestSession testSession = getTestSession();
+        if (testSession == null) {
+            throw new IllegalStateException("Internal error. Test session not initialized.");
```
sebady: Could we provide a more informative message here?

client/src/main/java/com/paypal/selion/internal/platform/grid/SeleniumGridListener.java
mach6: hmm.. yea. I'll refactor this a bit

client/src/main/java/com/paypal/selion/platform/grid/Grid.java
mach6: Hmm.. I guess we can say something like `Make sure the SeLion SeleniumGridListener is enabled and you have annotated your test with @MobileTest or @WebTest`...

PR #342, opened 2016-11-02
Fix #306: Custom elements in page containers

### LINKED ISSUE #306: SeLion-Code-Generator generates custom element incorrectly when defined in a Container
I am experiencing weird behavior when a custom HTML element is defined within a Container of a PageYAML file.  The resulting generated code incorrectly generates the getter for that element at the Page level instead of the Container level.  The object mapping for the element exists within the Container mapping (as expected).  Calling the getter at the Page level (as currently generated) results in a null locator being passed to the custom element, making that element unusable in the current implementation.  What configuration is needed for it to generate within the Container?

sebady:

---

- Fix #306
- Support custom elements in page object containers
- Also unrelated change in parent POM to remove
  eclipse lifecycle mapping plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -160,35 +160,6 @@
     <build>
         <pluginManagement>
             <plugins>
-                <!-- this plugin is a maven plugin designed to hold meta data to communicate with eclipse m2e plugin when 
-                    it processes a maven pom.xml. Without this, children of this pom file will get the warning 
-                    "maven-enforcer-plugin (goal enforce) is ignored by m2e" because there is no connector that tells 
-                    m2e how to integrate the maven plugin into eclipse. -->
-                <plugin>
-                    <groupId>org.eclipse.m2e</groupId>
-                    <artifactId>lifecycle-mapping</artifactId>
-                    <version>1.0.0</version>
-                    <configuration>
-                        <lifecycleMappingMetadata>
-                            <pluginExecutions>
-                                <pluginExecution>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -160,35 +160,6 @@
     <build>
         <pluginManagement>
             <plugins>
-                <!-- this plugin is a maven plugin designed to hold meta data to communicate with eclipse m2e plugin when 
-                    it processes a maven pom.xml. Without this, children of this pom file will get the warning 
-                    "maven-enforcer-plugin (goal enforce) is ignored by m2e" because there is no connector that tells 
-                    m2e how to integrate the maven plugin into eclipse. -->
-                <plugin>
```
mach6, line 167: Not needed anymore?

mach6:
Is there a test that can be added to ensure the bug doesn't resurface?

pom.xml
sebady: Recent m2e versions don't need this anymore and I don't believe any recent version of eclipse _requires_ this now. I know for certain eclipse (neon) builds SeLion fine w/o this.  For users still on old versions of Eclipse (for some reason) the recommendation is to add this to ~/.m2/settings.xml

mach6: good to learn and good catch!

mach6 (approved):
This looks good. Just need the test(s).

mach6:
merged this to 1.2.0

PR #354, opened 2017-05-15
Refactor codegen and add a cli-only version of it

mach6:
- Existing 'codegen' directory becomes codegen cli
- New 'codegen-maven' directory depends on cli variant and adds maven
  plugin specifics
- Refactor several classes to minimize code duplication
- Add a few tests
- Bump maven-plugin dependencies in parent pom
- Unrelated typo correction in 'client' AbstractElement.java

 ---

---

__Example usage__
```shell
$ java -jar target/SeLion-Code-Generator-CLI-2.0.0-SNAPSHOT-jar-with-dependencies.jar -h
Usage: <main class> [options]
  Options:
    -androidCustomElements
       Comma separated list of "android" custom elements to be included during
       code generation. For example: bar.foo.A,bar.foo.B
       Default: []
    -baseFolder
       Represents the base folder used for reading page asset files such as
       PageYAML.
       Default: GUIData
    -basePackage
       Represents the base package used for generated java classes.
       Default: com.paypal.selion.testcomponents
    -excludeDomains
       Comma separated list of "domains" to exclude during code generation. For
       example: foo,bar,baz
       Default: []
    -generatedSourcesDir
       The generated source directory to output to.
       Default: target/generated-test-sources
    -htmlCustomElements
       Comma separated list of "html" custom elements to be included during code
       generation. For example: bar.foo.A,bar.foo.B
       Default: []
    -iosCustomElements
       Comma separated list of "ios" custom elements to be included during code
       generation. For example: bar.foo.A,bar.foo.B
       Default: []
    -mobileCustomElements
       Comma separated list of "mobile" custom elements to be included during
       code generation. For example: bar.foo.A,bar.foo.B
       Default: []
    -resourcesDir
       The resources directory which contains the -baseFolder.
       Default: src/test/resources
    -sourceDir
       The source directory for classes which might depend on generated code
       (E.g. {PageObject}Ext classes)
       Default: src/test/java
    -workingDir
       The working directory.
       Default: ./
```


mach6:
@sebady @sundaramrajendran Please review.

codegen-maven/src/main/java/com/paypal/selion/plugins/CodeGeneratorMojoLogger.java
```diff
@@ -0,0 +1,127 @@
...
+\*-------------------------------------------------------------------------------------------------------------------*/
+
+package com.paypal.selion.plugins;
+
+import com.google.common.annotations.VisibleForTesting;
+import org.apache.maven.plugin.logging.Log;
+import org.apache.maven.plugin.logging.SystemStreamLog;
+
+public class CodeGeneratorMojoLogger implements CodeGeneratorLogger, Log {
+    private final Log instance;
```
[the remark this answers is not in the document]
mach6: won't fix

mach6:
@sebady ping on this

codegen-maven/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-core</artifactId>
+            <version>3.3.9</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.codehaus.plexus</groupId>
+                    <artifactId>plexus-classworlds</artifactId>
+                </exclusion>
```
sebady: Can you add a comment for these exclusions?

codegen-maven/src/main/java/com/paypal/selion/plugins/CodeGeneratorMojoLogger.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public CodeGeneratorMojoLogger() {
+        this.instance = new SystemStreamLog();
+    }
+
+    public CodeGeneratorMojoLogger(Log instance) {
+        this.instance = instance;
+    }
```
sebady, line 31: I'm not sure I see the value add in this class.  Can't we just use a Log instance everywhere CodeGeneratorMojoLogger is used?

sebady (approved):
Can we put the example usage above in a markdown file so it will be readily available for users of codegen CLI.

codegen-maven/pom.xml
mach6: sure... it's because of a maven enforcer conflict, fyi

codegen-maven/src/main/java/com/paypal/selion/plugins/CodeGeneratorMojoLogger.java
mach6: The reason the class `CodeGeneratorMojoLogger` exists is because `Log` is an interface in a maven plugin artifact dependency. I didn't want the SeLion Code Generator CLI to have any maven artifact dependencies. So, the CLI version defines the interface `CodeGeneratorLogger` ...

This class bridges `CodeGeneratorLogger` and `Log` so  the SeLion Code Generator maven plugin can have a`Log` and the code it depends on (SeLion Code Generator CLI) can have a `CodeGeneratorLogger` -- both as the same logger instance.

Issue #356, opened 2017-05-17
Better support for pages with asynchronous requests.


ibugaienko:
Hi, we are running into some issues with page validation being executed prematurely, would you please consider the following scenario:
Given 3 pages,
1. Login page (loginPage)
2. Premium user home page (premiumHomePage)
3. Standard user home page (homePage)

Suppose a user needs to provide their email/password and click a loginButton in loginPage, after which an asynchronous request is sent to a server which responds with a page to redirect to (premiumHomePage, homePage). While the browser is waiting for the response, some animation is shown to the user and the page remains the same (loginPage). If we were to use the following code: 

`loginPage.getLoginButton().clickAndExpectOneOf(loginPage, premiumHomePage, homePage)
`

Validation starts immediately after click() takes place, and since loginPage is among expected page objects (in case of server error for example) the above code will always return loginPage disregarding the actual landing page. The problem lies in the way that selenium handles asynchronous requests, as in the page is not loading so it will continue running the page validators.

For pages expressing this kind dynamic behavior it is desirable to forbid validation until the page finishes loading/updating. This would allow us work within clickAndExpect paradigm in cases like described above.

Solution Proposals:

1. Add semantics for expressing negation in pageValidators list, e.g. new actions: `isNotVisible`, `isNotPresent` or `!someElementLabel`, `!animationSpinner.isPresent`;
_or_
2. Add another ListPropertyType to YamlPage format, e.g. loadingValidators and logic checking for these before the actual validation starts.

Please let me know what you think of this, and then I could work on it if approved.

mach6:
@ibugaienko Go for it! While it is not clear to me which approach makes the most sense -- though 1 seems less complicated -- I see the issue you are trying to address here and welcome a PR.

ibugaienko:
@mach6 Thank you for quick response. I have both solutions in my fork on separate branches. If you do not mind I could raise a PR for each so that you could choose which one looks most appropriate to you. 
Also, please kindly note that this is the first time for me to make a contribution to an open source project so I would greatly appreciate to hear any feedback from you should you find anything requiring rework or improvement.

mach6:
@ibugaienko Not a problem. Yes, please raise a PR for each. We can review them both. In terms of contribution guidelines please review - https://github.com/paypal/SeLion#contributing if you have not already. In addition please make sure that the code does not contain any use of the `TAB` character (use four spaces instead) and has a righthand page margin of 120 characters.

mach6:
[...]

PR #357, opened 2017-05-31
Implements #356 (b) Added pageLoadingValidators and loading checks in click methods

ibugaienko:

---
This pull request addresses problem discussed in #356 (2nd solution)

### DIFF
--- a/client/src/main/java/com/paypal/selion/platform/html/AbstractElement.java
+++ b/client/src/main/java/com/paypal/selion/platform/html/AbstractElement.java
@@ -485,6 +485,9 @@ public abstract class AbstractElement implements Clickable, Hoverable {
         if (expected == null || expected.length == 0) {
             return;
         }
+        if (parent != null) {
+            WebDriverWaitUtils.waitUntilPageIsLoaded(parent.getCurrentPage());
+        }
         validatePresenceOfAlert();
         try {
             for (Object expect : expected) {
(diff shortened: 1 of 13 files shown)


mach6:
@ibugaienko code aside (the code looks good), It's not clear to me what the difference is between a `pageValidator` and `pageLoadingValidator`. When would a user use on over the other?

mach6:
@ibugaienko After looking closer at the code and the re-reading #356, it seems your intent is to have `pageLoadingValidators` process before any `pageValidators` does. Furthermore, like `pageValidators`, `pageLoadingValidators` appear to be completely optional. :)

Do you think `pageLoadingValidators` should also support negation (like your `pageValidators` PR)?

client/src/main/java/com/paypal/selion/platform/html/AbstractElement.java
```diff
@@ -485,6 +485,9 @@ public void click(Object... expected) {
         if (expected == null || expected.length == 0) {
             return;
         }
+        if (parent != null) {
+            WebDriverWaitUtils.waitUntilPageIsLoaded(parent.getCurrentPage());
+        }
```
mach6, line 490: Re: lines 510 to 512 in this this file, when `expected` contains a `WebPage` do you need to `waitUntilPageIsLoaded` on the `WegPage` before calling `waitUntilPageIsValidated`?

ibugaienko:
@mach6 , thank you for feedback!
You are right, in essence, `pageValidators` and `pageLoadingValidators` are same things, that is they are checklists pertaining to a particular page context, and both are optional. Former are used to identify pages, when all are `true`, latter indicate that page is loaded, when all are `false`.

To be honest, I can not think of a good example where negation would be needed in this case, but it could be added quite easily.

client/src/main/java/com/paypal/selion/platform/html/AbstractElement.java
ibugaienko: This is a very good question, I think you're right! In this case, loading check might be excessive. Should I move `waitUntilPageIsLoaded` inside the for loop and apply to first three if statements ?

mach6:
@ibugaienko  any update on this?

ibugaienko:
@mach6 , Hi, sorry for delay. I thought about your suggestion, but left code as it is. Here is why: there is no need to do loading checks if expected contains *only* another webpage, but if it is not the only expected object then the validation is still required, isn't it?

mach6:
@ibugaienko this is merged. Thanks!

PR #358, opened 2017-05-31
Implements #356 (a) Added support for negative pageValidators

ibugaienko:

---
This pull request addresses problem discussed in #356 (1st solution)


client/src/test/resources/GUIData/paypal/TestPageNegativeValidator.yml
```diff
@@ -0,0 +1,44 @@
...
+      containerButton:
+        locators:
+          US: ".//input"
+pageTitle:
+  US: "TestPage Generated By JavaScript"
+pageValidators: [
+  hiddenButton.isNotVisible,
+  "!hiddenButton.isVisible",
+  "!continueButton",
+  hiddenButton.is!Visible,
```
mach6: Don't think we need to support the `!` in the middle of these two (`is` and `Visible`) words. :) 

It just looks odd.

client/src/test/resources/GUIData/paypal/TestPageNegativeValidator.yml
```diff
@@ -0,0 +1,44 @@
...
+  hiddenButton.!isVisible,
+  hiddenButton.isVisibleNot,
```
mach6: Same feedback here.. `isVisiibleNot` just looks odd.. I don't think we need to support this.

client/src/test/resources/GUIData/paypal/TestPageNegativeValidator.yml
```diff
@@ -0,0 +1,44 @@
...
+  hiddenButton.isVisible!,
```
mach6: Same feedback here... The `!` on the end of `isVisible` looks odd. I don't think we need to support this

mach6:
@ibugaienko all in all I like this.. Just had the couple small points of feedback

client/src/test/resources/GUIData/paypal/TestPageNegativeValidator.yml
ibugaienko: hiddenButton is visible. NOT!
Looks a bit like a not-joke.

ibugaienko:
@mach6 , thank you, great feedback!
As the code written here, any of these will work, which is why I decided to include them all (even most odd ones) in the test. I will update the code checking `action` to be more picky.

mach6:
@ibugaienko  any update on this?

ibugaienko:
@mach6 , sorry to keep you waiting, I have updated the code, currently only below syntax is supported:
`hiddenButton.isNotVisible`
`"!hiddenButton.isVisible"`
`"!continueButton"`

Please let me know what you think.

PR #368, opened 2017-09-28
Bump appium java-client to 5.0.3

mach6:

---
Bump appium java-client to 5.0.3 

- Also updated the swipe and tap methods of SeLionAppiumAndroidDriver and
  SeLionAppiumIOSDriver to use TouchActions  -- both methods were removed
  from appium java-client
- Removed no-op rotation methods which now have implmentations in appium
  java-cient

### DIFF
--- a/appium-provider/src/main/java/com/paypal/selion/appium/platform/grid/SeLionAppiumAndroidDriver.java
+++ b/appium-provider/src/main/java/com/paypal/selion/appium/platform/grid/SeLionAppiumAndroidDriver.java
@@ -15,13 +15,14 @@
 
 package com.paypal.selion.appium.platform.grid;
 
+import io.appium.java_client.MultiTouchAction;
 import io.appium.java_client.TouchAction;
 import io.appium.java_client.android.AndroidDriver;
 
 import java.net.URL;
+import java.time.Duration;
 
 import org.openqa.selenium.Capabilities;
-import org.openqa.selenium.DeviceRotation;
 import org.openqa.selenium.Dimension;
 import org.openqa.selenium.Point;
 import org.openqa.selenium.WebElement;
@@ -34,8 +35,8 @@ import com.paypal.test.utilities.logging.SimpleLogger;
 
 /**
  * <code>SeLionAppiumAndroidDriver</code> provides facility to add custom {@link CommandExecutor} to
- * {@link AndroidDriver}. This class also implements the {@link SeLionAndroidBridgeDriver} interface to expose
- * methods for {@link UiObject} and its subclasses.
+ * {@link AndroidDriver}. This class also implements the {@link SeLionAndroidBridgeDriver} interface to expose methods
+ * for {@link UiObject} and its subclasses.
  */
 public class SeLionAppiumAndroidDriver extends AndroidDriver<WebElement> implements SeLionAndroidBridgeDriver {
 
@@ -54,7 +55,7 @@ public class SeLionAppiumAndroidDriver extends AndroidDriver<WebElement> impleme
     @Override
     public void click(WebElement webElement) {
         logger.entering(webElement);
-        this.tap(1, webElement, 1);
+        tap(1, webElement, 1);
         logger.exiting();
     }
 
@@ -162,7 +163,7 @@ public class SeLionAppiumAndroidDriver extends AndroidDriver<WebElement> impleme
     @Override
     public void longClick(WebElement webElement) {
         logger.entering(webElement);
-        this.tap(1, webElement, OPERATION_DURATION_MILLI_SECONDS);
+        tap(1, webElement, OPERATION_DURATION_MILLI_SECONDS);
         logger.exiting();
     }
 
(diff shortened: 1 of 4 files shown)


mach6:
@vikram1711 @elhuang can I also solicit your review of this change set?

appium-provider/src/main/java/com/paypal/selion/appium/platform/grid/SeLionAppiumAndroidDriver.java
```diff
@@ -245,27 +246,40 @@ public void clearText(WebElement webElement) {
...
-        return null;
+    public void tap(int fingers, int x, int y, int duration) {
+        MultiTouchAction multiTouch = new MultiTouchAction(this);
+
+        for (int i = 0; i < fingers; i++) {
+            TouchAction tap = new TouchAction(this);
+            multiTouch.add(tap.press(x, y).waitAction(Duration.ofMillis(duration)).release());
+        }
+        multiTouch.perform();
     }
```
vikram1711, line 283: Can you please confirm if it is multi finger tap or multi tap action or combination of both?
Based on my observation it is behaving like multi finger multi tap action.

Above implementation completely depends on the duration parameter for example if I want to perform a double tap with duration 200ms, this will become two long press actions, but if I perform a double tap action with 30ms then that will be treated as double tap. 

My suggestion would be not use raw actions rather use tap action exposed by TouchAction api-

Ex- 
```
MultiTouchAction multiTouch = new MultiTouchAction(driver);
        for (int i = 0; i < fingers; i++) {
            TouchAction action = new TouchAction(driver);
            multiTouch.add(action.tap(element));
        }
        multiTouch.perform();
```

vikram1711: Observation : Most of the methods are expecting WebElement as an argument, so in test first you have to grab the WebElement instance and then only you can invoke these methods. Why can't we follow the same approach we have taken for HTML elements where we are invoking the findElement internally to grab the element reference. I am not sure if that would violate the design of the framework.

appium-provider/src/main/java/com/paypal/selion/appium/platform/grid/SeLionAppiumAndroidDriver.java
```diff
@@ -245,27 +246,40 @@ public void clearText(WebElement webElement) {
...
     public void swipe(int startx, int starty, int endx, int endy) {
-        //super.swipe(startx, starty, endx, endy, OPERATION_DURATION_MILLI_SECONDS );
+        swipe(startx, starty, endx, endy, OPERATION_DURATION_MILLI_SECONDS);
+    }
 
-        // On Appium Android we mimic swipe via one finger tap
-        this.tap(1, starty, endx, OPERATION_DURATION_MILLI_SECONDS);
+    public void swipe(int startx, int starty, int endx, int endy, int duration) {
+        new TouchAction(this).press(startx, starty).waitAction(Duration.ofMillis(duration)).moveTo(endx, endy).release().perform();
     }
```
vikram1711, line 262: endx and endy - these are not absolute x,y value rather they represent the shift in x and y axis. So renaming them to shiftX and shiftY would make more sense

Also, consider a situation where you are swiping a carousel, most of the time you would be interested to go to a certain carousel content. To solve this problem if we can add a swipeCount argument then the author of the test can easily decide how many swipe is needed to go to the desired content, by default it will swipe only once.

appium-provider/src/main/java/com/paypal/selion/appium/platform/grid/SeLionAppiumIOSDriver.java
```diff
@@ -68,11 +70,8 @@ public WebElement findElementBy(By by) {
     @Override
     public void doubleTap(WebElement webElement) {
         logger.entering(webElement);
-        new TouchAction(this)
-                .tap(webElement).release()
-                .waitAction(DOUBLE_TAP_WAIT_TIME)
-                .tap(webElement).release()
-                .perform();
+        new TouchAction(this).tap(webElement).release().waitAction(Duration.ofMillis(DOUBLE_TAP_WAIT_TIME))
+                .tap(webElement).release().perform();
```
vikram1711, line 74: I would recommend to use appium provided doubleTap for iOS -
```
 new IOSTouchAction(this).doubleTap(webElement).perform();
```

appium-provider/src/main/java/com/paypal/selion/appium/platform/grid/SeLionAppiumIOSDriver.java
```diff
@@ -162,13 +161,27 @@ public String getValue(WebElement webElement) {
...
+        MultiTouchAction multiTouch = new MultiTouchAction(this);
+
+        for (int i = 0; i < fingers; i++) {
+            TouchAction tap = new TouchAction(this);
+            multiTouch.add(tap.press(element).waitAction(Duration.ofMillis(duration)).release());
+        }
+
+        multiTouch.perform();
+    }
+
```
vikram1711, line 186: I still need to test this action before I can comment on it since I am not sure if it is multi finger or multi tap action.

sebady (approved):
LGTM

appium-provider/src/main/java/com/paypal/selion/appium/platform/grid/SeLionAppiumIOSDriver.java
sebady: I believe these are all multi finger actions.

PR #370, opened 2017-10-10
cast to String exception when checking capabilities

alizelzele:
one of the capabilities is PLATFORM which the value is of type enum. checking for capabilities while PLATFORM capability is set will throw cast to string exception.

---

### DIFF
--- a/server/src/main/java/com/paypal/selion/grid/matchers/MinimalIOSCapabilityMatcher.java
+++ b/server/src/main/java/com/paypal/selion/grid/matchers/MinimalIOSCapabilityMatcher.java
@@ -1,5 +1,5 @@
 /*-------------------------------------------------------------------------------------------------------------------*\
-|  Copyright (C) 2015 PayPal                                                                                          |
+|  Copyright (C) 2015-2017 PayPal                                                                                          |
 |                                                                                                                     |
 |  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance     |
 |  with the License.                                                                                                  |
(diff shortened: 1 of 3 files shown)


mach6:
@alizelzele this looks good to me. The only question I have -- does appium support the `|` for specifying supported `platformVersion` on the node in a similar capacity?

mach6 (commented):
Waiting for answer to question.

alizelzele:
@mach6 appium will receive the correct platformVersion from request. the `|` will be used as a contract between node and hub to show how many `platformVersion`s are supported on that node.

for example on IOS node, we can have 9.2 and 9.3 available on the node, by specifying 9.2|9.3 in hub-config on node, we can tell hub that both of these platformVersions are available. which of them to be used is determined from the value specified in the request.

I also add another change. in new grid we are also considering default capability matcher values which some are enum in capability map. by using `.toString` instead of casting to String, they will be  converted to string correctly and it will prevent cast exception.

mach6:
@alizelzele Let me try asking a bit differently.

In the appium `--nodeConfig` file which is loaded by appium and registered with the hub, does appium support specifying `platformVersion` with a `|`?  

My understanding is when you would have `9.2` and `9.3` available on the node, you would setup the appium `--nodeConfig` to have  two separate entries in the `capabilities: [ ]` (array) portion of the file -- thus not requiring a `|` for the hub to do matching. 

In the Selenium web browser world, if I had two different versions of a browser installed (example firefox 50 and firefox 51) on the node, you would specify the two versions in the `-nodeConfig` file, in this fashion.

What am I missing? Can you show me an example `--nodeConfig` for appium that (1) does not work without your fix and (2) works with your fix.?

Thanks!

alizelzele:
@mach6 you are completely right. our nodes were configured with only 1 capability and I totally missed that it is actually an array. it works without any issue with multiple capability definition.

there is still enum casting to string issue though. I will update the pull request.

alizelzele:
updated the code and pull request.

mach6 (approved):
👍

mach6:
thanks @alizelzele

PR #375, opened 2017-11-18
support chrome-headless & firefox-headless -- option 1

mach6:
- [X] By placing an `X` in the preceding checkbox, I verify that I am a PayPal employee or
I have signed the [Contributor License Agreement](https://github.com/paypal/SeLion#contributing)

---

- support chrome-headless & firefox-headless
- deprecate phantomjs, selendroid, and ios-driver
- update travis build to use latest firefox for repo clones and sauce labs for parent repo
- update travis build to use chrome-headless and firefox-headless
- update travis build -- chrome tests are no longer allowed to fail

PR #376, opened 2017-12-05
Support for Chrome and Firefox headless -- option 2

mach6:

---
- Add support for headless chrome and firefox
- Deprecate 'client' support for phantomjs, selendroid, and ios-driver
- Update travis build to use firefox-esr for non-paypal repos and use sauce labs for the paypal one.


mach6:
@sebady This is another option instead of #375 .. does this appeal to you more?

mach6:
@sebady @sundaramrajendran please review as alternative implementation to PR #375

mach6:
Note: pending approval, I will squash all commits

sebady:
I think i prefer this implementation.  I do think we might consider adding a new field to WebTest class to indicate a test should be run headless.  (We already have browserHeight/browserWidth fields and `headless` is just another attribute of the browser value here).  Perhaps this would simplify/eliminate the new headless browser specific capability classes?

mach6:
@sebady Thanks. Yes, I thought about making it a new field on the `WebTest` annotation. The problem I have is ensuring the annotation value always wins. For example; given a `Config` value of `true` and a `WebTest` value of `false`, the annotation's value should win. If I define the default for the annotation field as `boolean headless() default false`, how do I know when the user provided `WebTest(headless = false)` or when the default was applied? In the former case I need to make the annotation win. In the later case I don't need to make the annotation win. If I don't define the field with a default value -- e.g.`boolean headless();`, it means a value must always be specified by the user -- which completely negates the `Config` value and would be a breaking change.

Because of messiness like this, I think I still prefer PR #375 -- but I am open to suggestions and not decided yet.

sebady:
Could we just have `WebTest(HeadLessOnly = false)` (default)  use global/local config  setting. `WebTest(headLessOnly = true)` always runs browser headless for test irrespective of the config settings.  I need to look closer but I thought this could work.

